PR #1, opened 2023-10-13
ci: add caching

OlivierDehaene:


Dantezxit:
services:
  tei-server:
    image: ghcr.io/huggingface/text-embeddings-inference:cpu-1.7
    ports:
      -

PR #70, opened 2023-11-13
feat: add `normalize` option

### LINKED ISSUE #69: Add Option to Disable Normalization in Result Vectors
### Feature request

Currently, the TEI framework automatically normalizes the result vectors upon retrieval. However, Sometimes It is needed to disable this behavior. To support this, introducing a new option that allows users to selectively enable or disable normalization would be beneficial.

### Motivation

The cross-encoder is a combination of an embedding layer and a classifier. I attempted to optimize only the embedding layer using TEI, but it wasn't successful due to normalization. In cases where one needs to optimize only a portion of the model, such as the embedding layer, using TEI, it would be beneficial to have the option to skip normalization.

### Your contribution

-

OlivierDehaene:
Close #69

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -60,15 +60,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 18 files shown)


OlivierDehaene:
Close #69

PR #73, opened 2023-11-13
add Jina Model class to handle Jina variant

JoanFM:
# What does this PR do?
This PR creates another type of model to handle Jina Models because I have observed that the way Jina models are run does not match the way they are supposed:

- [ ] They do not use `trust_remote_code` to initialize the model
- [ ] They do not use `mean_pooling` which is the way they are intended to be used.

I am open to discussions as if this should be generalized to another class not so targetted to a specific model (but some ALibiMeanPoolingModel) and also to discuss in the `get_model` method how to decide that this is going to be the model used.

* I open the PR as a draft because I still need to do testing but I would like to keep the discussion going

### DIFF
--- a/backends/python/server/text_embeddings_server/models/__init__.py
+++ b/backends/python/server/text_embeddings_server/models/__init__.py
@@ -6,8 +6,8 @@ from typing import Optional
 from transformers import AutoConfig
 from transformers.models.bert import BertConfig
 
-from text_embeddings_server.models.model import Model
 from text_embeddings_server.models.default_model import DefaultModel
+from text_embeddings_server.models.jina_model import JinaModel
 
 __all__ = ["Model"]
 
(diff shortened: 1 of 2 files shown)


JoanFM:
I will also update the code after https://github.com/huggingface/text-embeddings-inference/pull/70 to avoid normalizing

JoanFM:
> The jina model is only supported with the Candle backend. The Python backend is deprecated.
> 
> > I have observed that the way Jina models are run does not match the way they are supposed
> 
> Did you use the candle backend?

Interesting, I did not see, just assumed it was using that, I will take a look there, but yes I just passed the `model-id` and observed wrong results

JoanFM:
> > The jina model is only supported with the Candle backend. The Python backend is deprecated.
> > > I have observed that the way Jina models are run does not match the way they are supposed
> > 
> > 
> > Did you use the candle backend?
> 
> Interesting, I did not see, just assumed it was using that, I will take a look there, but yes I just passed the `model-id` and observed wrong results

Let me check if what I observe is just due to normalization

PR #80, opened 2023-11-16
feat: add Bloom* models

### LINKED ISSUE #32: Add support for Bloomz model
### Model description

Are there any plans to support the Bloomz models in the roadmap? It would be great if we could use TEI with our models [https://huggingface.co/cmarkea/bloomz-3b-embedding-ratio](https://huggingface.co/cmarkea/bloomz-3b-embedding-ratio)

However, this model requires pooling to take the last vector, which is not currently supported either.

### Open source status

- [X] The model implementation is available
- [X] The model weights are available

### Provide useful links for the implementation

_No response_

McPatate:
Adding support to the Bloom family of models

Fixes #32 

TODO:
- [ ] docs
- [ ] Tests

### DIFF
--- a/backends/candle/src/models/bloom.rs
+++ b/backends/candle/src/models/bloom.rs
@@ -1,8 +1,9 @@
-use crate::layers::LayerNorm;
-use candle::{DType, Device, Result, Tensor};
-use candle_nn::{Dropout, Embedding, Linear, VarBuilder};
+use candle::{DType, Device, IndexOp, Module, Result, Tensor};
+use candle_nn::{Dropout, Embedding, LayerNorm, Linear, VarBuilder};
 use serde::Deserialize;
-use text_embeddings_backend_core::Embedding;
+use text_embeddings_backend_core::{Batch, Pool};
+
+use crate::alibi::build_alibi_tensor;
 
 #[derive(Debug, Clone, PartialEq, Deserialize)]
 pub struct Config {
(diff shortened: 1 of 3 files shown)


backends/candle/src/models/bloom.rs
```diff
@@ -0,0 +1,245 @@
...
+    pub attention_softmax_in_fp32: bool,
+    pub seq_length: usize,
+    pub pad_token_id: usize,
+    pub masked_softmax_fusion: bool,
+    pub model_type: Option<String>,
+    pub n_embed: usize,
+    pub num_attention_heads: usize,
+}
+
+fn gelu_forward(x: &Tensor) -> Result<Tensor> {
```
LaurentMazare, line 33: A bit of a random comment but I think this is available directly in candle via `x.gelu()`.

McPatate: From the transformers docstring:
> Custom bias GELU function. Adapted from Megatron-DeepSpeed code.

The only code I could find in candle for the gelu function doesn't seem to be the same, but I'm not great at math so couldn't tell at a quick glance

LaurentMazare: I think the implementation is the same, see [here](https://github.com/huggingface/candle/blob/9ab3f9729fc1444687578a9dc913760b0d8d9963/candle-core/src/op.rs#L599), it's just a matter of calling `x.gelu` and would probably be a lot faster than doing the full thing as it would fuse all the ops together.

McPatate:
Closing for now as I won't be continuing the work here.

nolwennz:
Hello, do you still plan to integrate Bloom models? Thanks

PR #96, opened 2023-11-29
fix: fix tokenizers with both withespace and metaspace

### LINKED ISSUE #94: Difference in embedding when comparing SentenceTransformer vs TEI 0.4.0 and up for multilingual-e5-large
### System Info

Comparing SentenceTransformer output with TEI 0.2.2, 0.3.0, 0.4.0, and 0.5.0

### Information

- [x] Docker
- [ ] The CLI directly

### Tasks

- [ ] My own modifications

### Reproduction

I deployed the latest version of **multilingual-e5-large** on the following versions of TEI

1. v0.2.2
2. v0.3.0
3. v0.4.0
4. v0.5.0

They are ran using:
```
        # Replace with different versions
        image: ghcr.io/huggingface/text-embeddings-inference:89-0.3.0
        command: [ "text-embeddings-router" ]
        args:
          - "--port"
          - "7860"
          - "--dtype"
          - "float16"
          - "--model-id"
          - "/mnt/models/hub/models--intfloat--multilingual-e5-large/snapshots/9f78368af0062735ba99812349c562316e29f719"
```

I then generate an embedding using SentenceTransformer:
```
from sentence_transformers import SentenceTransformer
model = SentenceTransformer("intfloat/multilingual-e5-large", device='cpu')
payload = [' '.join(["Car"]*510)]
st_result = model.encode(payload)
```
and I send that same payload to the other 4 backends, enabling normalize and truncate where supported (although truncation is not happening).
I then calculate the cosine similarity between all these vectors:
```
from sklearn.metrics.pairwise import cosine_similarity
cos_sim = cosine_similarity(st_result, tei_result)
```

### Expected behavior

I expected the cosine similarities to be very close to 1 for all these values. However, I'm recording the following values:

st_result vs:
1. v0.2.2 = 0.9999881 (close enough)
2. v0.3.0 = 0.9999881 (close enough)
3. v0.4.0 = 0.95674335 (looks too far off)
4. v0.5.0 = 0.95674335 (looks too far off)

Is there a clear explanation of why this is happening? I did not a similar difference when using gte-large (ST vs TEI)

OlivierDehaene:
Close #94

### DIFF
--- a/router/src/main.rs
+++ b/router/src/main.rs
@@ -19,8 +19,8 @@ use text_embeddings_core::queue::Queue;
 use text_embeddings_core::tokenization::Tokenization;
 use text_embeddings_router::{ClassifierModel, EmbeddingModel, Info, ModelType};
 use tokenizers::decoders::metaspace::PrependScheme;
-use tokenizers::{PreTokenizerWrapper, Tokenizer};
 use tokenizers::pre_tokenizers::sequence::Sequence;
+use tokenizers::{PreTokenizerWrapper, Tokenizer};
 use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::util::SubscriberInitExt;
 use tracing_subscriber::{EnvFilter, Layer};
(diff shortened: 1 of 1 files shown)


ArthurZucker (approved):
LGTM, think we might have a similar issues with the `WhiteSpace` pretokenizer

PR #90, opened 2023-11-24
feat: add grpc router

OlivierDehaene:

PR #98, opened 2023-11-29
fix: enable http feature in http-builder

zhangfand:
#90 introduced regression to non-grpc build - server fails to start and complains "You must use one of `http` or `grpc`". This is due to the http feature flag is not turned on in the Dockerfile. This PR fixes that.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,7 +53,7 @@ COPY Cargo.lock ./
 
 FROM builder as http-builder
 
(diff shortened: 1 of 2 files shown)


zhangfand:
Hi @OlivierDehaene, I took the liberty to try fixing the issue myself. Can you please take a look when you have a chance?

OlivierDehaene (approved):
Thanks

PR #117, opened 2024-01-04
Update Dockerfile to install curl

jpbalarini:
# What does this PR do?

The base image doesn't have curl or wget installed, so we cannot do health checks when this is deployed as a service (for example, in AWS ECS, which is my use case).
I propose installing `curl` to be able to make container health checks like: `curl -f http://localhost:8080/health || exit 1`.

Thanks!

## Who can review?

@OlivierDehaene

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -80,6 +80,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-ins
     libomp-dev \
     ca-certificates \
     libssl-dev \
+    curl \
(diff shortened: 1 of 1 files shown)


karmi:
I've just ran into this issue myself, wondering why all containers are unhealthy despite running and serving... My code for healthcheck is:

```
healthcheck:
    test: curl --max-time 120 --retry 120 --retry-delay 1 --show-error --silent http://localhost/health
```

Can you please add `curl`, so it's possible to do a true healthcheck?

PR #120, opened 2024-01-04
feat: add support for Metal

### LINKED ISSUE #18: M chip Metal GPU support
### Feature request

I might be missing something, but on the M chip based macbook is it possible to force the crate to use the GPU instead of the CPU (the GPU would be much faster for generating embeddings).

### Motivation

We are  running a rust binary locally on the user device, and often they use macbooks, the tokenization is part of the slowest step. I would love to use the M chip GPU where possible.

### Your contribution

Happy to put in PRs to make this happen 🪄

OlivierDehaene:
close #18

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ classification models. TEI enables high-performance extraction for the most popu
 Ember, GTE and E5. TEI implements many features such as:
 
 * No model graph compilation step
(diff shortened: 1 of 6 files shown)


theskcd:
Thank you! 🛳️

PR #183, opened 2024-03-04
feat: support vertex api endpoint

drbh:
This PR adds a new endpoint for vertex support and conditionally includes openapi docs for this endpoint when built with the `google` feature


philschmid (commented):
~Why do we need a whole new route? 
Can't we use the regular "predict" route and just change the /path when AIP_X env are provided?~

My bad we need a new route since the Vertex Request has the `instances` object again.

router/src/http/server.rs
```diff
@@ -1101,6 +1219,8 @@ pub async fn run(
         .route("/tokenize", post(tokenize))
         // OpenAI compat route
         .route("/embeddings", post(openai_embed))
+        // Vertex compat route
+        .route("/vertex", post(vertex_compatibility))
```
philschmid, line 1223: This should not be needed since we only need the endpoint when /env is set.

drbh: in the TGI case we also always include the endpoint and only conditionally add the api docs. Should we avoid adding it in both cases if the feature is not on?

PR #188, opened 2024-03-06
bugfix: allow for larger request payload to avoid 413 when unwarranted

skeptrunedev:
# What does this PR do?

This PR enlarges the allowed request payload to be 0.5GB expressed in bytes - `67110000`. 

I am submitting this PR to fix a production bug that we were facing with the service where large re-rank requests would fail due to a 413 from axum. 

@OlivierDehaene OR @Narsil

### DIFF
--- a/router/src/http/server.rs
+++ b/router/src/http/server.rs
@@ -11,7 +11,7 @@ use crate::{
 };
 use ::http::HeaderMap;
 use anyhow::Context;
-use axum::extract::Extension;
+use axum::extract::{DefaultBodyLimit, Extension};
 use axum::http::HeaderValue;
 use axum::http::{Method, StatusCode};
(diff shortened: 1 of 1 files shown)


OlivierDehaene:
What's the batch size you are running?

gerritvd:
I was looking into making this configurable. What would be the preferred method of doing so? cmd line arg or env var?

skeptrunedev:
> I was looking into making this configurable. What would be the preferred method of doing so? cmd line arg or env var?

Env var preferably

skeptrunedev:
> What's the batch size you are running?

I don't recall exactly, but large enough to get this error I guess :sweat_smile:

OlivierDehaene:
Closed in favour of #210

PR #192, opened 2024-03-07
feat: add all methods to vertex API

OlivierDehaene:
@drbh what do you think?

@philschmid FYI this will modify the API to:

```
curl 127.0.0.1:3000/vertex \
    -X POST \
    -d '{"instances": [{"type": "embed", "inputs": "I like you."}]}' \
    -H 'Content-Type: application/json'
    
 # [{"type":"embed","result":[[-0.057003036,-0.022593308, ...]] }]
```

### DIFF
--- a/Dockerfile-cuda-all
+++ b/Dockerfile-cuda-all
@@ -33,6 +33,7 @@ FROM base-builder AS builder
 
 ARG GIT_SHA
 ARG DOCKER_LABEL
+ARG VERTEX
 
 # sccache specific variables
 ARG ACTIONS_CACHE_URL
@@ -45,7 +46,12 @@ COPY --from=planner /usr/src/recipe.json recipe.json
 
 FROM builder as builder-75
 
-RUN CUDA_COMPUTE_CAP=75 cargo chef cook --release --features candle-cuda-turing --no-default-features --recipe-path recipe.json && sccache -s
+RUN if [ $VERTEX = "true" ]; \
+    then \
+      CUDA_COMPUTE_CAP=75 cargo chef cook --release --features google --features candle-cuda-turing --no-default-features --recipe-path recipe.json && sccache -s; \
+    else \
+      CUDA_COMPUTE_CAP=75 cargo chef cook --release --features candle-cuda-turing --no-default-features --recipe-path recipe.json && sccache -s; \
+    fi;
 
 COPY backends backends
 COPY core core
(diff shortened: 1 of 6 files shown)


philschmid:
We should not make it different from what people would get on GKE? I think we should rather make it configurable what "type" is used via env, e.g. INFERENCE_TYPE="embed", "rank",.... We will have the same question for sagemaker too.

OlivierDehaene:
Then how do you use the multiple routes? You just decide a start time what's the route you want to serve? I think that's an inferior solution in every aspect.

OlivierDehaene:
`type` here allow you to use all TEI functionalities with a single route. `embed`, `embed_all`, `embed_sparse`, `tokenize`...

drbh:
@OlivierDehaene approved and looks good but deferring to @philschmid on the best interface

drbh (dismissed):
lgtm

OlivierDehaene:
Not my fault all the clouds can freaking figure out that oh surprise you may want to have more than a single route on your service...

philschmid:
> Then how do you use the multiple routes? You just decide a start time what's the route you want to serve? I think that's an inferior solution in every aspect.

You deploy 1 model, so you define 1 task, similar to other models, like `token-classification`. If you want to change, change your deployment I don't think that's an issue for companies.
It's way more confusing if payloads differen between solutions and make sure we document those correctly. I would not advise having a different request for Vertex AI and GKE. Thats not a good UX. Especially since the UX on Google Cloud to deploy is almost identical. 

The single route is something all Cloud ML Services currently have with Vertex and SageMaker.

OlivierDehaene:
> You deploy 1 model, so you define 1 task, similar to other models, like token-classification. If you want to change, change your deployment I don't think that's an issue for companies.

Except that's not the case. Some users use both `/embed` and `/embed_all` with one deployment. A lot of users use the method for the deployment and `/tokenize` at the same time.

> I would not advise having a different request for Vertex AI and GKE

Is it not already the case? Does GKE use the /vertex route? If not they don't have the `instances` shenanigans (@drbh is it mandatory to wrap the payload in `instances`?).

> The single route is something all Cloud ML Services currently have with Vertex and SageMaker.

Yeah and that's a poor design decision that everybody now need to live with.

philschmid:
> Does GKE use the /vertex route? 

No GKE can use all existing stuff. There is no `instances`. I commented in the [other PR that we don't need any ](https://github.com/huggingface/text-embeddings-inference/pull/183#discussion_r1512972220)`/vertex` route at all and just map the right /route, e.g. /embed when we are on Vertex AI.

OlivierDehaene:
What? https://github.com/huggingface/text-embeddings-inference/pull/183#pullrequestreview-1917272734

OlivierDehaene:
So if I understood you correctly:

- GKE can use the container as is
- Vertex deployments cannot because we need to add `instances` and `predictions` to the payloads. 

So WHATEVER we do we will have different APIs between the two.
Given the above, I think it is better to allow user the ability to use all routes with an additional type than just mapping arbitrarily a route.

OlivierDehaene:
@philschmid

philschmid:
> GKE can use the container as is

Yes 
> Vertex deployments cannot because we need to add instances and predictions to the payloads.

Yes 

>Given the above, I think it is better to allow user the ability to use all routes with an additional type than just mapping arbitrarily a route.

I would not do this since it would be different from what we did in TGI. Even if they have a different payload its the "same" just put into `instances` adding a "type" would make it more different and complex. I rather would define the "type" when a model is deployed.

OlivierDehaene:
What about we do both? 
Add `type` through env var to choose the default route.
If a user specifies `type` in the payload, it overrides the default.

This allow power user to have access to everything they want but you still have the default which has the same API.

This API is allowed by the OneOf union API in OpenAPI.

PR #260, opened 2024-05-05
Add optional CORS allow any option value in http server cli

kir-gadjello:
This PR adds option value '*' to --cors-allow-origin cli option to allow browser-based apps to use the embedding server directly. This is useful for local deployments of the embeddings inference server.


OlivierDehaene (approved):
Thanks

srjchauhan:
Hi,

When i specify * value in --cors-allow-origin i am still facing cors error in browser can you please help me with this

PR #263, opened 2024-05-12
Update `HUGGING_FACE_HUB_TOKEN` to `HF_API_TOKEN` in README 

kevinhu:
Updates the name of the token in the README from `HUGGING_FACE_HUB_TOKEN` to `HF_API_TOKEN` to avoid confusion.

The former is not used.


OlivierDehaene (approved):
Thanks

PR #293, opened 2024-06-19
ROCm support

fxmarty:
As per title. Support AMD GPUs through TEI backend.

For now, only embedding model with cls/mean pooling is tested, and tests need to actually be implemented.

MI210/MI250/MI300 can dispatch on CK flash attention 2, but other GPUs will default to manual attention implem (or SDPA). Only bert looks to be supported in the python backend.


fxmarty:
moved to https://github.com/huggingface/text-embeddings-inference/pull/295

PR #331, opened 2024-07-08
feat: add `/similarity` route

OlivierDehaene:
Required for compatibility.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3742,6 +3742,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fa42c91313f1d05da9b26f267f931cf178d4aba455b4c4622dd7355eb80c6640"
 
 [[package]]
+name = "simsimd"
+version = "4.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


cordawyn:
@OlivierDehaene Would it make sense to make it compatible with huggingface-hub client: https://github.com/huggingface/huggingface_hub/blob/ca111fb51193968faed548a4b865c79736822aed/src/huggingface_hub/inference/_client.py#L1416 ? It looks like they expect a bit different names for the attributes of `SimilarityInput` (https://github.com/huggingface/text-embeddings-inference/blob/4cc38bda85b2ecdbf34082796ca9738466483b85/router/src/http/types.rs#L364).

OlivierDehaene:
What do you mean? It is compatible.

OlivierDehaene:
The payload is: https://github.com/huggingface/huggingface_hub/blob/ca111fb51193968faed548a4b865c79736822aed/src/huggingface_hub/inference/_client.py#L1457

which is what we expect here.

cordawyn:
Oh, my apologies. I must've looked at a different revision or something. Thanks for the quick reply!

PR #338, opened 2024-07-10
feature: support bm42 embeddings

### LINKED ISSUE #337: Support for Qdrant bm42 document encoder
### Feature request

Support for https://huggingface.co/Qdrant/all_miniLM_L6_v2_with_attentions model's document embeddings

### Motivation

We want to test out bm42's performance when placed in a server when compared to splade-v2.

### Your contribution

I have a PR that proposes a way to add support for the sparse embedding model

cdxker:
# What does this PR do?

This pr adds support for `Qdrant/all_miniLM_L6_v2_with_attentions` I followed https://github.com/Anush008/bm42-rs as a reference and added a new pooling method of `bm42`.

Tested with running with command 

```sh
HF_TOKEN=hf_********************************** cargo run --no-default-features -F http -F ort -- --model-id Qdrant/all_miniLM_L6_v2_with_attentions --pooling bm42 --port 5888
```

Fixes #337

## Before submitting
- [X] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#start-contributing-pull-requests),
      Pull Request section?
      to it if that's the case. Issue #337 

## Who can review?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2293,6 +2293,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "defc4c55412d89136f966bbb339008b474350e5e6e78d2714439c386b3137a03"
 
 [[package]]
+name = "murmur3"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9252111cf132ba0929b6f8e030cac2a24b507f3a4d6db6fb2896f27b354c714b"
+
+[[package]]
 name = "native-tls"
 version = "0.2.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3477,6 +3483,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "rust-stemmers"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e46a2036019fdb888131db7a4c847a1063a7493f971ed94ea82c67eada63ca54"
+dependencies = [
+ "serde",
+ "serde_derive",
+]
+
+[[package]]
 name = "rustc-demangle"
 version = "0.1.24"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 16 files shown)


OlivierDehaene:
BM42 is super controversial.
See https://x.com/Nils_Reimers/status/1809334134088622217
I'm on the side of waiting a bit to see how this evolve.

OlivierDehaene:
To me Qdrant came at as a prety shady actor here so I would take everything they publish with a big grain of salt.

cdxker:
> To me Qdrant came at as a prety shady actor here so I would take everything they publish with a big grain of salt.

Fair. We really just want support for it so we could test on our own datasets and verify how good/bad it actually is. I agree a lot of that launch was shady; especially the quickwit bench portion. 

Any feedback on the actual implementation? Because the router function `async fn embed_sparse` expects a dense vector then `sparsifies` it is making it super expensive to serve a lot of sparse embedding models. In the case of this model it uses a `murmur3` hash which has a very large key Space.

Which means that `Infer` has to create and send a very large dense vector across a channel.  https://github.com/devflowinc/text-embeddings-inference/blob/dc62b26f04394c44aa3469c10ba8880d3076e087/backends/ort/src/bm42.rs#L293-L296

https://github.com/devflowinc/text-embeddings-inference/blob/dc62b26f04394c44aa3469c10ba8880d3076e087/core/src/infer.rs#L193-L204

This will impact adding support for other sparse models as well as this model. It would be a lot better if `Infer::embed_sparse` either:
1) Doesn't call the `embed` function and have each Backend have an `embed_sparse` function themselves.
2) Have `Infer::embed_spasrse` sparsify the vector before it gets sent back across the channel

I don't have enough context on what the best approach for this is, but it would open the opportunity for supporting other sparse embedding models that have larger keyspaces.

OlivierDehaene:
> Which means that Infer has to create and send a very large dense vector across a channel.

Why is that a problem? Moving a Vec shouldn't be a concern, it's not expensive as far as I know.
But yes, having a dedicated `embed_sparse` function in the backend trait could be interesting.

cdxker:
Creating a Vector with a max of 9223372036854776000 f32's is pretty expensive. In this case, even if they are all 0 filled

OlivierDehaene:
> Creating a Vector with a max of 9223372036854776000 f32's is pretty expensive. In this case, even if they are all 0 filled

Ok so when you say "Which means that Infer has to create and send a very large dense vector across a channel." you are only referring to the creation part? As I said the moving part ("send accross a channel") is not an issue.

PR #390, opened 2024-08-27
add missing link to TEI Gaudi in docs

rbrugaro:
Added link to TEI gaudi in docs. 

I followed same approach as in TGI documentation: https://huggingface.co/docs/text-generation-inference/index


yao-matrix:
we are mering tei-gaudi back to tei, so maybe we can hold on the doc thing as of now.

rbrugaro:
will upgrade docs if needed when tei-gaudi merged to tei

PR #489, opened 2025-02-07
Hpu bucketing

kaixuanliu:


backends/python/server/text_embeddings_server/utils/flash_attn.py
```diff
@@ -13,6 +13,7 @@
 is_hpu = is_hpu()
 use_ipex = use_ipex()
 
+
```
regisss, line 16: No need for this

kaixuanliu: You mean no need for is_hpu and use_ipex? The reason we use global variables here is to avoid calling `is_hpu()` function and `use_ipex()` function multiple times, as it will cost a lot of time for these two functions.

regisss: No no, I was talking about the blank line that was added. It's the only change to this file so I don't think we need it. I know it's not a big change, but let's try to keep PR as minimal as possible.

Narsil (approved):

PR #513, opened 2025-03-14
fix bug for `MaskedLanguageModel` class`

kaixuanliu:


kaixuanliu:
When using models like `sentence-transformers/all-mpnet-base-v2` to do embedding work, we should use `AutoModel` instead of `AutoModelForMaskedLM` to load the model , I made an adjustment in this PR, only when using splade pooling , we use `AutoModelForMaskedLM`. Pls help review, @regisss @Narsil

kaixuanliu:
cc @pi314ever

pi314ever:
Thanks @kaixuanliu for the fix. LGTM

pi314ever (approved):

backends/python/server/text_embeddings_server/models/types.py
```diff
@@ -1,4 +1,5 @@
 import os
+import math
```
regisss, line 2: Do we need this?

kaixuanliu: Yes, it is an omitted import in https://github.com/huggingface/text-embeddings-inference/pull/489, we need add this.

regisss (approved):
LGTM

regisss:
cc @Narsil

PR #515, opened 2025-03-14
make a WA in case Bert model do not have `safetensor` file

kaixuanliu:
@regisss @Narsil pls help review

### DIFF
--- a/backends/python/server/text_embeddings_server/models/__init__.py
+++ b/backends/python/server/text_embeddings_server/models/__init__.py
@@ -67,8 +67,13 @@ def get_model(model_path: Path, dtype: Optional[str], pool: str):
                 )
             try:
                 return FlashBert(model_path, device, datatype)
-            except:
-                return DefaultModel(model_path, device, datatype, pool, trust_remote=TRUST_REMOTE_CODE)
+            except FileNotFoundError as e:
+                logger.info(
+                    "Do not have safetensors file for this model, use default transformers model path instead"
+                )
+                return DefaultModel(
+                    model_path, device, datatype, pool, trust_remote=TRUST_REMOTE_CODE
+                )
         if config.architectures[0].endswith("Classification"):
             return ClassificationModel(
                 model_path, device, datatype, trust_remote=TRUST_REMOTE_CODE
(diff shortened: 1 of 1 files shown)


backends/python/server/text_embeddings_server/models/__init__.py
```diff
@@ -65,7 +65,10 @@ def get_model(model_path: Path, dtype: Optional[str], pool: str):
                 return DefaultModel(
                     model_path, device, datatype, pool, trust_remote=TRUST_REMOTE_CODE
                 )
-            return FlashBert(model_path, device, datatype)
+            try:
+                return FlashBert(model_path, device, datatype)
+            except:
```
regisss, line 70: Maybe we can catch the specific error linked to no safetensors file rather than catching everything?

kaixuanliu: I think the main purpose is to make the model run smoothly first, hence no matter whether it is an error caused by no safetensor file or sth else, we can use `DefaultModel` path to do the backup choice. WDYT?

regisss: The problem here is that we risk missing completely unrelated errors, which is not good.
Thinking about it more, I don't get why `DefaultModel` would work in this case and not `FlashBert`. Can you send a command to reproduce it and the logs of the error please?

kaixuanliu: `DefaultModel` path uses transformers modeling while `FlashBert` implements modeling itself, and it reads weight file from safetensors [L305](https://github.com/huggingface/text-embeddings-inference/blob/main/backends/python/server/text_embeddings_server/models/flash_bert.py#L305),  for some models they do not have this file, which will cause the error. 
Cmd line to reproduce:
`docker run -p 8080:80 -v $volume:/data --network host --runtime=habana -e HABANA_VISIBLE_DEVICES=all -e MAX_WARMUP_SEQUENCE_LENGTH=512 -e OMPI_MCA_btl_vader_single_copy_mechanism=none --cap-add=sys_nice --ipc=host tei_hpu --model-id BAAI/bge-base-zh-v1.5 --dtype bfloat16 --auto-truncate --pooling cls`

kaixuanliu: I have updated the `try ... except ...` part.

kaixuanliu:
@regisss Hi, I have updated the code, pls have a review.

Narsil:
LGTM Thanks for this contribution, I'll probably follow up by raising the first error in case of failure (since it's likely to be the real root cause).

Not using `safetensors` is considered a bug at this point given the magnitude of the security implementations.

PR #517, opened 2025-03-14
add docker build for python backend in CI workflow

kaixuanliu:


baptistecolle:
See PR https://github.com/huggingface/text-embeddings-inference/pull/518

PR #526, opened 2025-03-24
Refine model file download for python backend

kaixuanliu:
For models like [jinaai/jina-embeddings-v2-base-code](https://hf.co/jinaai/jina-embeddings-v2-base-code), Rust side will download 2 separate directories `models--jinaai--jina-embeddings-v2-base-code/` and `models--jinaai--jina-bert-v2-qk-post-norm/`, while existing implementation only pass 1 Path to python backend. This PR refines model file download so that we can run models below using python backend:

- [Alibaba-NLP/gte-Qwen2-7B-instruct](https://hf.co/Alibaba-NLP/gte-Qwen2-7B-instruct)
- [Alibaba-NLP/gte-Qwen2-1.5B-instruct](https://hf.co/Alibaba-NLP/gte-Qwen2-1.5B-instruct)
- [jinaai/jina-embeddings-v2-base-code](https://hf.co/jinaai/jina-embeddings-v2-base-code)


kaixuanliu:
@regisss @Narsil pls help review

Narsil:
I'm sorry but I do not understand, how does it downloads 2 models ?  I cannot reproduce on my side.

Can you provide steps to reproduce, it seems the candle backend properly downloads only 1 model.

Propagating `model_id` is not something desirable a priori, the fact that the model id is resolved first, and then only a path is used makes debugging usually much simpler.

Narsil:
I can now reproduce. The trust-remote-code behavior of this model is suprising to me, looking into what's going on.

Narsil:
Thanks for surfacing this, discussing internally we figured there are some security implications of this behavior, which we're most likely going to close, so this behavior will go away (and force every repo to own their own remote code, so that trust_remote_code cannot be abused as much).

Before we're going through with this, we're trying to figure out the implications, are you aware of other models with that behavior ?

kaixuanliu:
I have not met other models with this unexpected behavior. But these three I listed above are in TEI README part. Our customers are asking support for these 3 models:
[Alibaba-NLP/gte-Qwen2-7B-instruct](https://hf.co/Alibaba-NLP/gte-Qwen2-7B-instruct)
[Alibaba-NLP/gte-Qwen2-1.5B-instruct](https://hf.co/Alibaba-NLP/gte-Qwen2-1.5B-instruct)
[jinaai/jina-embeddings-v2-base-code](https://hf.co/jinaai/jina-embeddings-v2-base-code)

kaixuanliu:
Will change to another safer manner to support these 3 models

PR #546, opened 2025-03-28
Make `sliding_window` for `Qwen2` optional

alvarobartt:
# What does this PR do?

This PR makes the `sliding_window` to be `Option<usize>` instead of `usize` as it can be set to `null`; additionally, the `model` prefix is only pushed for `FlashQwen2` if available, as the `Qwen2` model does not come with the prefix since it's an instance of `Qwen2Model` not `Qwen2ModelForCausalLM`, meaning that it doesn't have neither the `lm_head` nor the `model` prefix.

This PR then enables models as e.g. https://huggingface.co/mims-harvard/ToolRAG-T1-GTE-Qwen2-1.5B (also Azure asked us to add support for this model and a recent discussion has been opened at https://huggingface.co/mims-harvard/ToolRAG-T1-GTE-Qwen2-1.5B/discussions/1, which is unrelated as they are trying to use TGI instead of TEI, but that would unlock serving this model).

## Before submitting

- [x] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#start-contributing-pull-requests), Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).

## Who can review?

@Narsil (and @pagezyhf for context on the Azure support)

### DIFF
--- a/backends/candle/src/models/flash_qwen2.rs
+++ b/backends/candle/src/models/flash_qwen2.rs
@@ -21,8 +21,8 @@ struct Qwen2Attention {
 
 impl Qwen2Attention {
     pub fn load(vb: VarBuilder, config: &Qwen2Config) -> Result<Self> {
-        if config.sliding_window.is_some() {
-            candle::bail!("Sliding window is not supported for Qwen2");
+        if config.use_sliding_window {
+            candle::bail!("Sliding window is not supported for Qwen2",);
         }
 
         let num_attention_heads = config.num_attention_heads;
--- a/backends/candle/src/models/flash_qwen2.rs
+++ b/backends/candle/src/models/flash_qwen2.rs
@@ -264,7 +264,15 @@ impl FlashQwen2Model {
             ModelType::Embedding(pool) => pool,
         };
 
-        let vb = vb.pp("model");
+        // Pushing the prefix for `model` is apparently only required if the model architecture is
+        // ForCausalLM as it contains the `lm_head`, other than that, the `model` key won't be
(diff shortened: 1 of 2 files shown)


alvarobartt:
@Narsil not sure if the `contains_tensor` approach is the best towards preventing pushing the prefix `model` when not there i.e. when the Transformers model instance is `Qwen2Model` instead of `Qwen2ModelForCausalLM` (not sure this also affects `Mistral` but I'll check and if that's the case submit a follow-up PR).

Narsil (approved):
We should update the non-flash version too I think.

Narsil:
Checking for tensor name is not the best, but honestly there's no great way. The least worst is to know deterministically based on some other information (so the error checking works well).

alvarobartt:
> Checking for tensor name is not the best, but honestly there's no great way. The least worst is to know deterministically based on some other information (so the error checking works well).

Yes I agree with you, couldn't find of a better way, it seems that's tight to the `arch` in this case being `Qwen2Model`, so we could check the arch and if different that `Qwen2Model` then calling `pp(model)` before passing it to the `Qwen2Attention`? That may look cleaner, but still we would need to "manipulate" the VarBuilder in advance.

PR #459, opened 2024-12-25
Implement the `ModernBert` model

kozistr:
# What does this PR do?

Close #457

- upgrade the `tokenizer` crate from `0.19.1` to `0.21.0` to address a `ModernBert` tokenizer issue.
- implement `ModernBert` model
    * it may work on CPU, CUDA (w/o FA), ~~and MPS~~.
    * `ModernBert` uses local attention. however, I'm unfamiliar with `candle_flash_attn` and don't have any GPU to test FA2 w/ local attn, so the `FlashModernBert` implementation remains unsupported at this time.
- implement a classification head for `ModernBert`
- tested
  - `nomic-ai/modernbert-embed-base`
  - `answerdotai/ModernBert-base/large`

### Log

```
$ ./target/release/text-embeddings-router --model-id ./ModernBERT-base/ --port 8888 --pooling cls --dtype float32
2024-12-25T07:34:46.753673Z  INFO text_embeddings_router: router/src/main.rs:175: Args { model_id: "./Mod*******-*ase/", revision: None, tokenization_workers: None, dtype: Some(Float32), pooling: Some(Cls), max_concurrent_requests: 512, max_batch_tokens: 16384, max_batch_requests: None, max_client_batch_size: 32, auto_truncate: false, default_prompt_name: None, default_prompt: None, hf_api_token: None, hostname: "0.0.0.0", port: 8888, uds_path: "/tmp/text-embeddings-inference-server", huggingface_hub_cache: None, payload_limit: 2000000, api_key: None, json_output: false, otlp_endpoint: None, otlp_service_name: "text-embeddings-inference.server", cors_allow_origin: None }
2024-12-25T07:34:46.817444Z  WARN text_embeddings_router: router/src/lib.rs:184: Could not find a Sentence Transformers config
2024-12-25T07:34:46.817472Z  INFO text_embeddings_router: router/src/lib.rs:188: Maximum number of tokens per request: 8192
2024-12-25T07:34:46.817622Z  INFO text_embeddings_core::tokenization: core/src/tokenization.rs:28: Starting 8 tokenization workers
2024-12-25T07:34:46.883933Z  INFO text_embeddings_router: router/src/lib.rs:230: Starting model backend
2024-12-25T07:34:46.884247Z  INFO text_embeddings_backend_candle: backends/candle/src/lib.rs:239: Starting ModernBert model on Cpu
[...]

PR #561, opened 2025-04-05
Update the logic for local attetion mask in ModernBERT

kozistr:
# What does this PR do?

In the previous PR #459, `local_attention` can only be worked on `CPU` [due to the `abs()` operation](https://github.com/huggingface/text-embeddings-inference/pull/459#issuecomment-2779101344).

So, I've made a change to calculate the `window mask` in pure Rust and then create the Tensor from it. Maybe, this could make `ModernBERT` run on both MPS and CUDA too.

I've checked the output with [this script](https://github.com/huggingface/text-embeddings-inference/pull/459#issuecomment-2764840553) and it seems identical to before.

- [x] tested on CPU (local machine, WSL2)
- [x] tested on T4 GPU (Kaggle notebook, TEI server runs successfully)
- [ ] tested on MPS (might work, but needs to be tested)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@Narsil, @alvarobartt, @ivarflakstad

### DIFF
--- a/backends/candle/src/models/modernbert.rs
+++ b/backends/candle/src/models/modernbert.rs
@@ -577,11 +577,10 @@ impl ModernBertModel {
         Ok(extended_attention_mask)
     }
 
-    fn get_local_attention_mask(&self, attention_mask: &Tensor) -> Result<Tensor> {
-        let (_, _, seq_len, _) = attention_mask.dims4()?;
+    fn get_window_mask(&self, seq_len: usize) -> Result<Tensor> {
         let window_size = self.local_attention / 2;
 
-        let mask: Vec<f32> = (0..seq_len)
+        let inverted_window_mask: Vec<f32> = (0..seq_len)
             .flat_map(|i| {
                 (0..seq_len).map(move |j| {
                     if (j as i32 - i as i32).abs() > window_size as i32 {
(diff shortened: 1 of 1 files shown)


backends/candle/src/models/modernbert.rs
```diff
@@ -577,28 +577,49 @@ impl ModernBertModel {
...
-            .unsqueeze(0)?
-            .unsqueeze(0)?
-            .broadcast_as(mask_shape)?;
+        let seq_len = global_attention_mask.dim(2)?;
+        let window_mask = self.get_window_mask(seq_len)?;
+        let window_mask = (window_mask * min_value)?.to_dtype(self.dtype)?;
 
-        let zero_tensor = Tensor::zeros_like(&attention_mask)?;
-        let local_attention_mask = attention_mask.where_cond(&window_mask, &zero_tensor)?;
+        let local_attention_mask = global_attention_mask.broadcast_add(&window_mask)?;
```
kozistr, line 620: I just added them instead of using masked_fill (or where_cond), which is referred to [here](https://github.com/huggingface/candle/blob/main/candle-transformers/src/models/modernbert.rs#L226).

When values from both masks (`global_attention_mask` and `window_mask`) are `f32::MIN`, the output becomes f32::MIN + f32::MIN, and it could overflow. I honestly haven't looked into it in detail, so we need to check that it's okay.

Narsil:
@kozistr I think everything already works on MPS, it's been checked by @ErikKaum https://github.com/huggingface/text-embeddings-inference/pull/562

During the flash fix I forced the mask creation in CPU candle (to enable non-flash CUDA execution), and that fixed metal as well.
I think keeping everything in candle instead of Rust, makes it slightly nicer, because we could switch to on-device execution if we simply created the missing kernels (which is relatively easy to do).

No matter what, having a mask that is only created in the first part of the model is acceptable imho.

Narsil:
That being said, we could definitely clean up a little bit that code imho.

The HashMap<bool, XX> could be replaced with a simple `struct Mask{local: Tensor, global: Tensor}` which should increase readability (and the overhead of a HashMap is not negligible even though I doubt it can be measure in this particular instance).

kozistr:
> @kozistr I think everything already works on MPS, it's been checked by @ErikKaum #562
> 
> During the flash fix I forced the mask creation in CPU candle (to enable non-flash CUDA execution), and that fixed metal as well. I think keeping everything in candle instead of Rust, makes it slightly nicer, because we could switch to on-device execution if we simply created the missing kernels (which is relatively easy to do).
> 
> No matter what, having a mask that is only created in the first part of the model is acceptable IMHO.

I thought there were still some issues on MPS, but I missed that PR. Thanks for checking on that.
And I agree with your point that keeping everything in Candle would be great!

> That being said, we could definitely clean up a little bit that code imho.
> 
> The HashMap<bool, XX> could be replaced with a simple `struct Mask{local: Tensor, global: Tensor}` which should increase readability (and the overhead of a HashMap is not negligible even though I doubt it can be measure in this particular instance).

I second this. We could refactor the mask part with `struct` instead of `HashMap` regarding readability.


Then, I'm gonna close this PR and make another contribution later!
 
Thanks for your time to review this PR :)

Narsil:
Thanks a lot to you.

PR #574, opened 2025-04-09
[Docs] Update quick tour

NielsRogge:
# What does this PR do?

This PR updates the quick tour to include examples using the `InferenceClient` and `OpenAI` Python SDKs besides `cURL`. 

To do:

- [ ] I assume `InferenceClient` is only supported for the embed endpoint, not reranking or sequence classification? cc @Wauplin 
- [ ] verify code snippets


Narsil (approved):

Wauplin (commented):
Thanks! 

> I assume InferenceClient is only supported for the embed endpoint, not reranking or sequence classification? cc

FYI `InferenceClient.feature_extraction` is for /embed and `InferenceClient.sentence_similarity` is for /similarity (but I don't think we want to document it). We don't have anything yet for reranking indeed.

PR #575, opened 2025-04-09
optimize the performance of FlashBert Path for HPU

kaixuanliu:
Use `WhereIsAI/UAE-Large-V1` model to do  benchmark, below is the throughput(seq/s) comparison:
| bs | before | after |
|-----|-----|-----|
| 1 | 199.65 | 219.61 |
| 2 | 239.61 | 284.78 |
| 4 | 321.37 | 373.13 |
| 8 | 506.40 | 549.61 |
| 16 | 759.07 | 822.17 |
| 32 | 1028.31 | 1285.57 |
| 64 | 1130.67 | 1708.73 |
| 128 | OOM | 1030.06 |

### DIFF
--- a/backends/python/server/text_embeddings_server/models/flash_bert.py
+++ b/backends/python/server/text_embeddings_server/models/flash_bert.py
@@ -323,19 +323,21 @@ class FlashBert(Model):
     def embed(self, batch: Union[FlashBatch, PaddedBatch]) -> List[Embedding]:
         if isinstance(batch, PaddedBatch):
             input_lens = batch.attention_mask.cumsum(-1)[:, -1].to(torch.int32)
-            max_input_lens = input_lens.max().item()
+            max_input_lens = 0  # This value will not be used
             cu_seqlens = torch.cat(
                 (input_lens.new_tensor([0]), input_lens.cumsum(-1).int())
             )
             mask = batch.attention_mask.bool()
-            batch_size = input_lens.size(0)
+            bsz, tgt_len = mask.size()
+            min_val = torch.finfo(self.dtype).min
             attn_mask = torch.full(
-                [batch_size, 1, 1, mask.shape[-1]],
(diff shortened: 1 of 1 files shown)


kaixuanliu:
@regisss @Narsil pls help review, thx!

backends/python/server/text_embeddings_server/models/flash_bert.py
```diff
@@ -323,19 +323,21 @@ def batch_type(self) -> Union[FlashBatch, PaddedBatch]:
     def embed(self, batch: Union[FlashBatch, PaddedBatch]) -> List[Embedding]:
         if isinstance(batch, PaddedBatch):
             input_lens = batch.attention_mask.cumsum(-1)[:, -1].to(torch.int32)
-            max_input_lens = input_lens.max().item()
+            max_input_lens = 0  # This value will not be used
```
Narsil, line 326:
```suggestion
```

NIT

Narsil (approved):
LGTM

regisss (approved):

kaixuanliu:
@regisss @Narsil Hi, can you help merge this PR?

backends/python/server/text_embeddings_server/models/flash_bert.py
regisss: I guess there are cases where the forward of the model does need a right value for this right? Otherwise why not removing it there?

kaixuanliu: Well, this is a common file shared by CPU/XPU andd HPU devices. On CPU/XPU, we do need this variable with exact meaning, while on HPU, we do not have real `varlen_attention` API, so we pass attn_mask to replace its functionality. Here we just need to set a random value for `max_input_lens`. This line cannot be deleted, as we need to pass it to L352.

PR #595, opened 2025-04-18
fix xpu env issue that cannot find right libur_loader.so.0

kaixuanliu:


Dockerfile-intel
```diff
@@ -115,7 +115,7 @@ RUN wget -qO - https://repositories.intel.com/gpu/intel-graphics.key | gpg --dea
 RUN wget -O- https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB \
 | gpg --dearmor | tee /usr/share/keyrings/oneapi-archive-keyring.gpg > /dev/null && echo "deb [signed-by=/usr/share/keyrings/oneapi-archive-keyring.gpg] https://apt.repos.intel.com/oneapi all main" | tee /etc/apt/sources.list.d/oneAPI.list
 
-RUN apt-get update && apt install -y intel-basekit cmake python3-dev ninja-build pciutils
+RUN apt-get update && apt install -y intel-basekit cmake vim python3-dev ninja-build pciutils
```
regisss, line 118: Not a big deal, but why do we need to install vim here?

kaixuanliu: Well, it is not necessary, but sometimes I need to do some debug work directly in the container, it's convenient to add it here.

regisss (approved):
LGTM

regisss:
cc @Narsil

PR #599, opened 2025-04-23
remove optimum-habana dependency

kaixuanliu:
We are going to do modeling in tei instead of depends on optimum-habana for HPU.


regisss (approved):
LGTM

regisss:
cc @Narsil

Issue #615, opened 2025-05-27
ModernBert reranker - Different results from vs transformers library


kwnath:
## Description
I'm very green to this field so apologies in advance if I ask some stupid questions or make some assumptions that are incorrect.
___

I have a Modernbert reranker initially run on the `Transformers` library and saw that `TEI` would perform inference faster so I migrated them over. What I observed was different outputs, initially I thought this was fine but the differences are quite large so I stepped into the different processes from tokenisation to pooling strategy and the activation function. What I found was two areas where the two libraries differ that produce a significant difference between the two libraries ouputs:

1. Tokenisation - Modernbert on TEI does not add additional padding (maybe this is fine?)
2. The pooling strategy used by default is CLS, where as in `Transformers` it uses the value in `config.json`

I tested both libraries on the same machine where TEI was run based on the Dockerfile-cuda. One small note is that I'm running TEI on `float16` (I didn't have enough memory to run with float32) but I don't think this would drastically change the results.

I made a simple high level example:

```python
query = "How do I sell my shirt?"
texts = [
    "You can sell your shirt by going to the sell page and clicking the sell button.",
    "Ketchup is a condiment that is made from tomatoes.",
    "You can sell your apple in the store.",
    "How you can sell your clothes online."
]
pairs = [[query, text] for text in texts]


# Transformers: 
[{'index': 0, 'score': 0.99937457},
 {'index': 3, 'score': 0.82623243},
 {'index': 2, 'score': 0.0025561687},
 {'index': 1, 'score': 1.2188445e-05}]

# TEI:
[{'index': 0, 'score': 0.9993761},
 {'index': 3, 'score': 0.29017562},
 {'index': 2, 'score': 0.0047737225},
 {'index': 1, 'score': 1.2219076e-05}]
```

### Model
The model that i'm running: https://huggingface.co/tomaarsen/reranker-ModernBERT-large-gooaq-bce

```
{
  "model_id": "../models/toma-reranker",
  "model_sha": null,
  "model_dtype": "float16",
  "model_type": {
    "reranker": {
      "id2label": {
        "0": "LABEL_0"
      },
      "label2id": {
        "LABEL_0": 0
      }
    }
  },
  "max_concurrent_requests": 512,
  "max_input_length": 8192,
  "max_batch_tokens": 16384,
  "max_batch_requests": null,
  "max_client_batch_size": 32,
  "auto_truncate": false,
  "tokenization_workers": 32,
  "version": "1.7.0",
[...]

PR #616, opened 2025-05-27
Add mean pooling strategy for Modernbert classifier

kwnath:
# What does this PR do?

Apologies in advance, new this this area and Rust 😅.

I noticed differences between Transformers and TEI libraries when running rerankers with a `mean` classifier pooling strategy. [More details here](https://github.com/huggingface/text-embeddings-inference/issues/615).

Fixes https://github.com/huggingface/text-embeddings-inference/issues/615

## Before submitting
- [x] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
      Pull Request section?
      to it if that's the case.

## Who can review?
Not sure who might be best to review 😄  @OlivierDehaene OR @Narsil

### DIFF
--- a/backends/candle/src/models/flash_modernbert.rs
+++ b/backends/candle/src/models/flash_modernbert.rs
@@ -260,7 +260,15 @@ impl FlashModernBertModel {
 
         let (pool, classifier) = match model_type {
             ModelType::Classifier => {
-                let pool = Pool::Cls;
+                let pool = if let Some(pooling) = &config.classifier_pooling {
+                    match pooling.as_str() {
+                        "cls" => Pool::Cls,
+                        "mean" => Pool::Mean,
+                        _ => Pool::Cls,
+                    }
+                } else {
+                    Pool::Cls
+                };
 
                 let classifier: Box<dyn ClassificationHead + Send> =
                     Box::new(ModernBertClassificationHead::load(vb.clone(), config)?);
--- a/backends/candle/src/models/flash_modernbert.rs
+++ b/backends/candle/src/models/flash_modernbert.rs
@@ -260,15 +260,11 @@ impl FlashModernBertModel {
 
         let (pool, classifier) = match model_type {
             ModelType::Classifier => {
-                let pool = if let Some(pooling) = &config.classifier_pooling {
-                    match pooling.as_str() {
-                        "cls" => Pool::Cls,
-                        "mean" => Pool::Mean,
-                        _ => Pool::Cls,
-                    }
-                } else {
-                    Pool::Cls
-                };
+                let pool: Pool = config
+                    .classifier_pooling
+                    .as_deref()
+                    .and_then(|s| Pool::from_str(s).ok())
+                    .unwrap_or(Pool::Cls);
 
                 let classifier: Box<dyn ClassificationHead + Send> =
                     Box::new(ModernBertClassificationHead::load(vb.clone(), config)?);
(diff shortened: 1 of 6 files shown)


backends/candle/tests/test_modernbert.rs
```diff
@@ -202,3 +202,61 @@ fn test_modernbert_classification() -> Result<()> {
 
     Ok(())
 }
+
+#[test]
```
kwnath, line 206: Having a bit of trouble running this test and not super pleased with this test but it;s just to test that both cls/mean produce different outputs as expected.

kozistr: How about just testing with the [reranker-ModernBERT-large-gooaq-bce](https://huggingface.co/tomaarsen/reranker-ModernBERT-large-gooaq-bce) model directly, which is a mean classifier-pooling, instead of `gte-reranker` model with mean pooling?

I think we can check whether classifier pooling correctly works by running the `reranker-ModernBERT-gooqa-bce` model, while `cls` classifier pooling has already been verified in the above test! (implicitly assume cls and mean will behave differently)

> unable to download the HF model, hitting a 429, 

when I faced that issue, in my case, I manually download the model via git or huggingface-cli on my local, and then set `HUGGINGFACE_HUB_CACHE` to point to that local path.

backends/candle/src/models/flash_modernbert.rs
```diff
@@ -260,7 +260,15 @@ impl FlashModernBertModel {
 
         let (pool, classifier) = match model_type {
             ModelType::Classifier => {
-                let pool = Pool::Cls;
+                let pool = if let Some(pooling) = &config.classifier_pooling {
```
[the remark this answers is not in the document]
kozistr: I'm also not confident about this, maybe we could consider `std::str::FromStr` for the `Pool` enum or simply add a `from_str` like below. It would be great if the maintainers could provide some guidance or feedback on this :)

```
impl std::str::FromStr for Pool {

    fn from_str(s: &str) -> Result<Self, Err> {
        match s {
            "cls" => Ok(Pool::Cls),
            "mean" => Ok(Pool::Mean),
            "splade" => Ok(Pool::Splade),
            "last_token" => Ok(Pool::LastToken),
            _ => Err(()),
        }
    }
}
```

```
impl Pool {
    pub fn from_option_str(pool: Option<&str>) -> Pool {
        match pool {
            Some("cls") => Pool::Cls,
            Some("mean") => Pool::Mean,
            Some("splade") => Pool::Splade,
            Some("last_token") => Pool::LastToken,
            _ => Pool::Cls, // Default fallback
        }
    }
}
```

kozistr (commented):
i'm just a passerby lol. I leaved some comments, to hope if it could help you in some way :) thanks

backends/candle/tests/test_modernbert.rs
kwnath: It couldn't find from the cache but I managed to just export a hugginface token, all good!

```shell
successes:
    test_modernbert_classification_mean_pooling

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 3 filtered out; finished in 3.94s
```

kozistr (approved):
looks good to me!

Narsil:
Thanks a lot for this ! 
This PR is good, I will merge as-is and instantly modify it to remove `FromStr` into using purely `serde` to parse the pooling directly.

This will allow for better validation (instead of defaulting to Cls if we can't read the config which is a footgun, it's better to error out).
And it will remove most of the code added here :)

PR #635, opened 2025-06-13
chore: `Dockerfile-cuda` - Retain major CC when pruning static cuBLAS lib

polarathene:
# What does this PR do?

Pruning cuBLAS for CC 7.5 now also retains `sm_70` in addition to the `sm_75` target. See https://github.com/huggingface/text-embeddings-inference/pull/610#issuecomment-2969143241 for more information.

### DIFF
--- a/Dockerfile-cuda
+++ b/Dockerfile-cuda
@@ -51,7 +51,9 @@ RUN --mount=type=secret,id=actions_results_url,env=ACTIONS_RESULTS_URL \
     --mount=type=secret,id=actions_runtime_token,env=ACTIONS_RUNTIME_TOKEN \
     if [ ${CUDA_COMPUTE_CAP} -ge 75 -a ${CUDA_COMPUTE_CAP} -lt 80 ]; \
     then  \
-        nvprune --generate-code code=sm_${CUDA_COMPUTE_CAP} /usr/local/cuda/lib64/libcublas_static.a -o /usr/local/cuda/lib64/libcublas_static.a; \
+        # When pruning cuBLAS for minor CC versions, ensure you retain the major CC version as well:
+        # https://docs.nvidia.com/cuda/cublas/#nvprune
+        nvprune --generate-code code=sm_70 --generate-code code=sm_${CUDA_COMPUTE_CAP} /usr/local/cuda/lib64/libcublas_static.a -o /usr/local/cuda/lib64/libcublas_static.a; \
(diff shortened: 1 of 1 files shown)


polarathene:
**NOTE:** There is no known need to do this for TEI, however [Nvidia encourages retaining the major CC and any minors in-between](https://docs.nvidia.com/cuda/cublas/#nvprune) when using `nvprune` on cuBLAS.

---

Feel free to close the PR if you prefer to avoid until there's a relevant bug report. My understanding is it should only be an issue when using a kernel from cuBLAS that would defer to `sm_70` when it'd have been equivalent for `sm_75`.

For example in the current base image used to build, `sm_70` has 184 cubins vs `sm_75` containing only 8:

```console
$ cuobjdump --list-elf /usr/local/cuda/lib64/libcublas_static.a | grep -oE '\.sm_70.*\.' | wc -l
184

$ cuobjdump --list-elf /usr/local/cuda/lib64/libcublas_static.a | grep -oE '\.sm_75.*\.' | wc -l
8

# Individual cubins:
$ cuobjdump --list-elf /usr/local/cuda/lib64/libcublas_static.a | grep -E '\.sm_75.*\.'
ELF file    5: libcublas_static.5.sm_75.cubin
ELF file   13: libcublas_static.13.sm_75.cubin
ELF file   21: libcublas_static.21.sm_75.cubin
ELF file   29: libcublas_static.29.sm_75.cubin
ELF file   37: libcublas_static.37.sm_75.cubin
ELF file   45: libcublas_static.45.sm_75.cubin
ELF file   53: libcublas_static.53.sm_75.cubin
ELF file   61: libcublas_static.61.sm_75.cubin
```

I'm not entirely sure why the minor CC versions in-between (when present) might matter to be retained.

---

The concern does not apply to the other two supported real archs handled via `nvprune` as `sm_80` is already provided, while `sm_90` does not target anything newer (_since it's the [only arch for that CC major](https://developer.nvidia.com/cuda-gpus)_):

https://github.com/huggingface/text-embeddings-inference/blob/53eae1b2d6091ee312fcd2421450df641be0c37e/Dockerfile-cuda#L57-L60

alvarobartt (changes requested):
Hey @polarathene whilst this is true, given that we don't officially support Volta i.e., CUDA compute capability 7.0; I don't see any reason why we should retain the minor `sm_` here being `sm_70` as opposed to only `sm_75` i.e., Turing.

PR #479, opened 2025-01-29
(FIX): CI Security Fix - branchname injection

glegendre01:
# What does this PR do?

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#start-contributing-pull-requests),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #184, opened 2024-03-05
feat: support vertex api endpoint

drbh:
This PR adds a new endpoint for vertex support and conditionally includes openapi docs for this endpoint when built with the `google` feature

reopened version of https://github.com/huggingface/text-embeddings-inference/pull/183

PR #631, opened 2025-06-11
Add `HiddenAct::Silu` (remove `serde` alias)

alvarobartt:
# What does this PR do?

TL;DR Leverage `HiddenAct.forward` and handle SiLU and SwiGLU separately

This PR fixes the `forward` method in the `HiddenAct` implementation to use the SiLU activation when the activation is SwiGLU, since SwiGLU is SiLU + down projection to half the size since it's split on the intermediate dimension (as per https://github.com/huggingface/candle/blob/17313a4226a6c6bde444d28b4be4f0f96d155be7/candle-nn/src/ops.rs#L44-L47), and since `candle` brings an implementation for it, then the `serde` alias for `Silu` has been removed to handle separately both Silu and Swiglu.

Additionally, to unify the codebase, this PR leverages the `forward` method in `HiddenAct` instead of the previous `match self.act`.

Fixes #629

## Before submitting

- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [X] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#start-contributing-pull-requests), Pull Request section?
- [X] Was this discussed/approved via a GitHub issue or the [forum](https://discuss.huggingface.co/)? Please add a link to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [ ] Did you write any new necessary tests?

## Who can review?

@kozistr since they recently introduced the `forward` method in `HiddenAct` and @Narsil to verify the current patch

PR #637, opened 2025-06-13
chore: Revise `Dockerfile-cuda-all`

polarathene:
# What does this PR do?

Less than the verbosity implies 😅 
- Leverage [HereDoc syntax](https://docs.docker.com/reference/dockerfile/#here-documents) for multi-line `RUN` instructions. This makes the content more akin to an inlined multi-line shell script, easier to grok.
- The upstream base images for the `base-builder` and `runtime-base` stages are now configurable `ARG` at the top of the `Dockerfile`. Users can more easily build with their preferred CUDA + Ubuntu bases, while also making it simpler to maintain updates in the `Dockerfile`.
- `ARG VERTEX` was hoisted up to the top as well. This is now only referenced by a `FROM` instruction, where two stages are available to match against the `true`/`false` value of this ARG. Doing so simplifies the `RUN` instructions, removing the repeated conditionals that only exist to add `google` as a feature.
  - This could probably be simplified further if you instead switch to a more generic ARG for customizing additional cargo features, then anyone building it could use `--build-arg WITH_EXTRA_FEATURES=google`, although I see [very little relevance within this project via a GH search for `vertex`](https://github.com/search?q=repo%3Ahuggingface%2Ftext-embeddings-inference%20VERTEX&type=code)?
  - [Support was added March 2024](https://github.com/huggingface/text-embeddings-inference/pull/184)
  - [Follow-up that added to `Dockerfile-cuda-all`](https://github.com/huggingface/text-embeddings-inference/pull/192) by @OlivierDehaene (_[No other `Dockerfile` variant appears to have this logic](https://github.com/search?q=repo%3Ahuggingface%2Ftext-embeddings-inference%20VERTEX&type=code)_)
  - [Not used by CI builds](https://github.com/huggingface/text-embeddings-inference/blob/53eae1b2d6091ee312fcd2421450df641be0c37e/.github/workflows/build.yaml#L130-L135) (_related [`matrix.json`](https://github.com/huggingface/text-embeddings-inference/blob/53eae1b2d6091ee312fcd2421450df641be0c37e/.github/workflows/matrix.json#L48-L55)_)
- `SSCACHE` + `CARGO_CHEF` ENV were changed to ARG with a `_RELEASE` suffix, then co-located to their equivalent `RUN`. No need for them to be ENV, co-location to their install logic seemed relevant, but they could be hoisted to the top of the stage if you prefer. Doing such invalidates all layers of the stage when those are version bumped however.
- Likewise ENV `RUSTC_WRAPPER` + `PATH` were co-located to the relevant rustup install.
- I've added a note for context about the `RUN` secrets being specific to sccache usage. However as the comment highlights, since [this Jan 2025](https://github.com/huggingface/text-embeddings-inference/pull/479) change the ARG `SCCACHE_GHA_ENABLED` seems redundant?
  - The `RUN` secrets will expect values provided, so it's a bit awkward to opt-out (_I haven't checked but assume there's no guidance to interested users_).
  - Additionally [this specific image build isn't using `sccache` in CI](https://github.com/huggingface/text-embeddings-inference/blob/53eae1b2d6091ee312fcd2421450df641be0c37e/.github/workflows/matrix.json#L48-L55).
  - Looking at recent CI runs for the other `Dockerfile` variants that do use `sccache`, [they're not leveraging any cache hits](https://github.com/huggingface/text-embeddings-inference/actions/runs/15620193833/job/44002874430#step:10:1298) from `sccache`. It seems to be broken / ineffective?

### DIFF
--- a/Dockerfile-cuda-all
+++ b/Dockerfile-cuda-all
@@ -1,43 +1,60 @@
-FROM nvidia/cuda:12.2.0-devel-ubuntu22.04 AS base-builder
+# syntax=docker/dockerfile:1
 
-ENV SCCACHE=0.10.0
-ENV RUSTC_WRAPPER=/usr/local/bin/sccache
-ENV PATH="/root/.cargo/bin:${PATH}"
-# aligned with `cargo-chef` version in `lukemathwalker/cargo-chef:latest-rust-1.85-bookworm`
-ENV CARGO_CHEF=0.1.71
+ARG BASE_IMAGE_BUILDER=nvidia/cuda:12.2.0-devel-ubuntu22.04
+ARG BASE_IMAGE_RUNTIME=nvidia/cuda:12.2.0-runtime-ubuntu22.04
+# Build with support for Google Cloud Vertex AI:
+# https://cloud.google.com/vertex-ai
+ARG VERTEX="false"
 
-RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
-    curl \
-    libssl-dev \
-    pkg-config \
-    && rm -rf /var/lib/apt/lists/*
+FROM $(BASE_IMAGE_BUILDER) AS base-builder
+ARG DEBIAN_FRONTEND=noninteractive
+RUN <<HEREDOC
+    apt-get update
+    apt-get install -y --no-install-recommends \
+        curl \
+        libssl-dev \
+        pkg-config
 
-# Donwload and configure sccache
-RUN curl -fsSL https://github.com/mozilla/sccache/releases/download/v$SCCACHE/sccache-v$SCCACHE-x86_64-unknown-linux-musl.tar.gz | tar -xzv --strip-components=1 -C /usr/local/bin sccache-v$SCCACHE-x86_64-unknown-linux-musl/sccache && \
+    rm -rf /var/lib/apt/lists/*
+HEREDOC
+
+# Download sccache
+ARG SCCACHE_RELEASE=0.10.0
+RUN <<HEREDOC
+    ASSET_NAME="sccache-v${SCCACHE_RELEASE}-x86_64-unknown-linux-musl"
+
+    curl -fsSL "https://github.com/mozilla/sccache/releases/download/v${SCCACHE_RELEASE}/${ASSET_NAME}.tar.gz" \
+      | tar -xz -C /usr/local/bin --strip-components=1 "${ASSET_NAME}/sccache"
+
     chmod +x /usr/local/bin/sccache
+HEREDOC
 
 RUN curl https://sh.rustup.rs -sSf | bash -s -- -y
-RUN cargo install cargo-chef --version $CARGO_CHEF --locked
+ENV PATH="/root/.cargo/bin:${PATH}"
+ENV RUSTC_WRAPPER=/usr/local/bin/sccache
+
+# aligned with `cargo-chef` version in `lukemathwalker/cargo-chef:latest-rust-1.85-bookworm`
+ARG CARGO_CHEF_RELEASE=0.1.71
+RUN cargo install cargo-chef --version ${CARGO_CHEF_RELEASE} --locked
+
 
 FROM base-builder AS planner
-
 WORKDIR /usr/src
-
 COPY backends backends
 COPY core core
 COPY router router
 COPY Cargo.toml ./
 COPY Cargo.lock ./
-
 RUN cargo chef prepare  --recipe-path recipe.json
 
-FROM base-builder AS builder
 
+FROM base-builder AS builder-setup
 ARG GIT_SHA
 ARG DOCKER_LABEL
-ARG VERTEX="false"
 
-# sccache specific variables
+# sccache supports GHA cache as storage backend:
+# https://github.com/mozilla/sccache/blob/main/docs/GHA.md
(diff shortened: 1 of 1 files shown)


polarathene:
## Feedback - Caching appears to be wasted complexity

Presumably these caching features never really worked for Docker image builds in the project.

**TL;DR:**
- Cargo chef seems redundant?
  - To benefit from image layer cache across builds, you need to avoid invalidating the cache, the use of `GIT_SHA` + `DOCKER_LABEL` ARGs prior to chef cook invalidates that layer each time a new commit is pushed to trigger a build.
  - Caching these layers to S3 is thus a waste of storage unless you're actually building an image for the same commit hash again where the is existing cache to use and no other inputs are changing that'd likewise invalidate a cached layer from being used (rare?).
- sccache seems redundant?
  - Your CI runs that use it have no cache hits, even for the same PRs with successive runs from new commits. Caches are made and uploaded to GHA cache, a PR can upload multiple 180MB artifacts, one from each run.
  - Reliant upon GHA cache by default despite ARG for opt-out, the RUN expect to receive secret inputs for GHA cache usage.

---

I've not used sccache myself, nor cargo chef, so during this refactor I tried to respect the separate `RUN` layers.

As it's now clear that the secrets mounted are for `sccache` and that cargo chef doesn't rely on the separate layers of each variant for caching, those could all be merged to minimize the noise of the secrets mounting, it's just questionable if that is really adding any value since having `sscache` with reliance upon GHA isn't as likely to be used in local builds either? I'm not quite sure I understand the intent when the caching feature clearly is not working as intended?

---

Looking at the [recent workflow runs](https://github.com/huggingface/text-embeddings-inference/actions/workflows/build.yaml):

> image

> image

We can see repeated runs are near an hour long. The ones that are not were canceled to begin a new run due to the concurrency setting:

https://github.com/huggingface/text-embeddings-inference/blob/53eae1b2d6091ee312fcd2421450df641be0c37e/.github/workflows/build.yaml#L42-L48

Take this [workflow run for example](https://github.com/huggingface/text-embeddings-inference/actions/runs/15582126164/job/43879718401#step:10:1466):

> image

That was the `cargo build` that came after a `cargo chef cook`, no cache utilized, nor does it appear that the chef hydration is working as intended as crates were downloaded and compiled again? (_if there's no change in the recipe, provided the layer was cached it should have been skipped as a cache hit for that RUN_)

What's actually being cached via the added complexity?

https://github.com/huggingface/text-embeddings-inference/blob/53eae1b2d6091ee312fcd2421450df641be0c37e/Dockerfile-cuda#L65-L91

For reference the cook + build times are 525s + 740s (_vs CPU variant: 120s + 240s_), as it stands you might as well just build without any attempt to cache.

Beyond that you have CI using Amazon S3 for caching image build layers (_with max mode, so it should capture intermediate stages too_):

https://github.com/huggingface/text-embeddings-inference/blob/53eae1b2d6091ee312fcd2421450df641be0c37e/.github/workflows/build.yaml#L141-L142

Looking at the [associated PR](https://github.com/huggingface/text-embeddings-inference/pull/631), nothing should be affecting the deps.

So perhaps something about your inputs is invalidating cached image layers?:

https://github.com/huggingface/text-embeddings-inference/blob/53eae1b2d6091ee312fcd2421450df641be0c37e/.github/workflows/build.yaml#L130-L138

I'm not sure why the `planner` stage doesn't appear cached, but it does seem like your `GIT_SHA` + `DOCKER_LABEL` ARGs are going to get changed with each commit, thus invalidating any layer caching there (_waste to upload any cache of that stage to S3_):

https://github.com/huggingface/text-embeddings-inference/blob/53eae1b2d6091ee312fcd2421450df641be0c37e/Dockerfile-cuda#L34-L38

These are used to store the build commit SHA (long + short) into the binary for some reason within the router crate (_introduced in the [fist public commit / code-drop](https://github.com/huggingface/text-embeddings-inference/commit/b26fd46686f1ed229b1966c2b0ef105d67d1a673) for the repo_):

https://github.com/huggingface/text-embeddings-inference/blob/53eae1b2d6091ee312fcd2421450df641be0c37e/router/build.rs#L5-L23

So that explains why each `cargo chef cook` wasn't helping as intended.

Over to [`sccache` GHA cache storage for this workflow where it is actually used](https://github.com/huggingface/text-embeddings-inference/actions/caches?query=sort%3Asize-desc), that same PR 631 (_and related 627_) was uploading cache, it just wasn't reusable:

> image

---

Seems these were implemented without verification that the features actually work. While I can understand wanting to include a SHA into your binary builds, how relevant is that actually for PR / CI beyond actual releases?

Notably within Docker images where this information can be instead encoded in the image metadata or as ENV/tag? (_pretty sure you can have `build.rs` retrieve this from `Cargo.toml` or similar instead_)

alvarobartt:
Hey @polarathene thanks a lot for your contribution, given the recent changes on both `Dockerfile-cuda` and `Dockerfile-cuda-all` as per https://github.com/huggingface/text-embeddings-inference/pull/699 (which removes both AWS SageMaker and Google Cloud Vertex AI from those Dockerfiles), https://github.com/huggingface/text-embeddings-inference/pull/831 (that bumps Ubuntu 24.04 and CUDA 12.9), among others.

Additionally, I'm not a fan of the HereDoc syntax as it doesn't align well with how we ship those Dockerfiles hence I'd prefer not to include those changes; anyway I'll investigate the issues with the `gh cache` and the `sccache` to ensure that we're benefiting from those.

Thanks again 🤗

PR #646, opened 2025-06-17
Fix Qwen3

### LINKED ISSUE #642: Qwen3-Embedding models: embeddings from TEI differ sharply from Sentence-Transformers reference
### System Info

image: ghcr.io/huggingface/text-embeddings-inference:hopper-sha-a69cc2e
hardware: H100/H200

### Information

- [x] Docker
- [ ] The CLI directly

### Tasks

- [ ] My own modifications

### Reproduction

```from sentence_transformers import SentenceTransformer
import requests

model = SentenceTransformer("Qwen/Qwen3-Embedding-0.6B")

txt = "this is a test"

st_embedding = model.encode([txt])

tei_embedding = requests.post(
    "http://localhost:8000/embed", #TEI server
    json={"inputs": [txt]},
)

np.dot(tei_embedding.json()[0], st_embedding[0])
```

### Expected behavior

For Qwen3 models, the vectors returned by TEI have very low cosine similarity to vectors produced with the same model loaded via Sentence-Transformers.
Using identical text, pooling mode, and normalization, the cosine similarity between TEI and ST is often <0.2.
When running the same test with another model, e.g. BAAI/bge-base-en-v1.5, the similarity is 1.0, as expected.

kozistr:
# What does this PR do?

Fixes #642

## Changes

- [x] left padding (w/ pad token id)
- [x] causal attention mask
- [x] fix last token pooling for the batch case

* Currently, batch inference works fine, but single-query inference has a problem.
* (updated) I've further investigated this issue and found that it's due to the attention logic.
    * `self.config._attn_implementation` of `Qwen3 Embedding` config is `sdpa` by default, and if we change this value to `eager`, which is naive impl of self-attention, we could check the identical output with this PR as expected.
    * but, looks like `sdpa` performs something different, need to investigate more on this

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@Narsil @alvarobartt


alvarobartt (approved):
Thanks for the PR @kozistr and apologies I missed those points! LGTM

Narsil (approved):

kozistr:
@alvarobartt @Narsil thanks for the quick check!

I've further investigated this issue (== single query inference seems weird) and found that it's due to the attention logic. `self.config._attn_implementation` of `Qwen3 Embedding` config is `sdpa` by default, and if we change this value to `eager`, which is naive implementation of self-attention, we could check the identical output with this PR as expected.

Seems like `sdpa` performs something different, need to investigate more on this.

I'll let you know when I found the root cause : ) thanks!

PR #656, opened 2025-06-25
Qwen3 hpu support

kaixuanliu:


kaixuanliu:
@regisss pls help review. We use our own scripts to do a benchmark, w/ this PR, for model Qwen3/Qwen3-Embedding-8B, the throughput improved from original 121.2 seq/s to 126.93 seq/s

PR #659, opened 2025-06-26
change HPU warmup logic: seq length should be with exponential growth

kaixuanliu:


kaixuanliu:
As model input length increases greatly (like Qwen3), we need to change HPU warmup lengths with exponential growth. @regisss pls help review. Thx

regisss:
cc @Narsil: in `backends/src/lib.rs`, only the `warmup_hpu` method is changed

PR #660, opened 2025-06-26
Add `Dense` layer for `2_Dense/` modules

alvarobartt:
# What does this PR do?

This PR adds support for `2_Dense/` modules, since some models as e.g. https://huggingface.co/sentence-transformers/LaBSE require the extra `Dense` module i.e., an extra `Linear` layer on top of the pooled embeddings, when generating the embeddings.

So on, this PR introduces the `DenseLayer` trait, impl for `Dense` and adds `DenseConfig`, which are models with basically a single `Linear` layer, pulling the configuration from `2_Dense/config.json` and the model weights from `2_Dense/model.safetensors`.

> [!NOTE]
> The `2_Dense/` is only required when generating embeddings, meaning that it will only apply to the `Embedding` model type, whereas the `Reranker` and `Classifier` are not affected by this addition, so on, neither the `rank` or `predict` methods for the given backend.

This PR solves the issue recently reported at https://discuss.huggingface.co/t/inference-result-not-aligned-with-local-version-of-same-model-and-revision/160514.

Additionally, this PR also fixes a shape mismatch issue produced when performing matrix multiplication of 2D tensors on Metal devices due to the `candle` Metal kernels expecting the tensors to be contiguous. It seems that the error only arises on Metal for 2D tensors, where as for e.g. 3D tensors it seems to be working just fine without having to use `.contiguous()` (which is expensive as it needs to clone the tensor).

## Reproduce

To ensure that the implementation was working fine and producing successful results i.e., `allclose` like checks are true, and the cosine similarity is 1.0 (or as close as possible), the following test has been run:

1. Deploy Text Embeddings Inference (TEI) as e.g.:

```bash
cargo run --release --features candle,http --no-default-features -- --model-id sentence-transformers/LaBSE --dtype float16
```

2. Then, once it's running run the following Python script (requires `torch`, `transformers`, `sentence-transformers`, `accelerate` and `numpy`):

```python
import numpy as np
import requests
from sentence_transformers import SentenceTransformer

model = SentenceTransformer(
    "sentence-transformers/LaBSE",
    model_kwargs={
        "torch_dtype": "float16",
        "device_map": "mps",
    },
)

out_py = model.encode(
    "What is Deep Learning?",
    normalize_embeddings=True,
    convert_to_numpy=True,
)

response = requests.post(
    "http://localhost:3000/embed",
    json={
        "inputs": "What is Deep Learning?",
        "normalize": True,
    },
)

response.raise_for_status()
out = response.json()[0]
out_http = np.array(out, dtype=np.float16)

print(f"Embeddings are close: {np.allclose(out_py, out_http, atol=1e-3, rtol=1e-4)=}")


def cosine_similarity(x: np.ndarray, y: np.ndarray) -> float:
    dot_product = np.dot(x, y)
    norm_x = np.linalg.norm(x)
    norm_y = np.linalg.norm(y)
    return dot_product / (norm_x * norm_y)


print(f"The similarity score is: {cosine_similarity(out_py, out_http)=}")
```

It should produce the following on any combination of device (CPU, MPS, CUDA) and dtype (float32, float16):

```
Embeddings are close: np.allclose(out_py, out_http, atol=1e-3, rtol=1e-4)=True
The similarity score is: cosine_similarity(out_py, out_http)=np.float16(1.0)
```

## Before submitting
- [X] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#start-contributing-pull-requests), Pull Request section?
- [X] Was this discussed/approved via a GitHub issue or the [forum](https://discuss.huggingface.co/)? Please add a link to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).

## Who can review?

@Narsil

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,16 @@ Options:
 
           [env: DEFAULT_PROMPT=]
 
+      --dense-path <DENSE_PATH>
+          Optionally, define the path to the Dense module required for some embedding models.
+
+          Some embedding models require an extra `Dense` module which contains a single Linear layer and an activation function. By default, those `Dense` modules are stored under the `2_Dense` directory, but there might be cases where different `Dense` modules are provided, to convert the pooled embeddings into different dimensions, available as `2_Dense_<dims>` e.g. https://huggingface.co/NovaSearch/stella_en_400M_v5.
+
+          Note that this argument is optional, only required to be set if the path to the `Dense` module is other than `2_Dense`. And it also applies when leveraging the `candle` backend.
+
+          [env: DENSE_PATH=]
+          [default: 2_Dense]
+
       --hf-token <HF_TOKEN>
           Your Hugging Face Hub token
 
@@ -304,10 +314,10 @@ Options:
 
           [env: CORS_ALLOW_ORIGIN=]
 
-      -h, --help
+  -h, --help
           Print help (see a summary with '-h')
 
-      -V, --version
+  -V, --version
           Print version
 ```
 
--- a/docs/source/en/cli_arguments.md
+++ b/docs/source/en/cli_arguments.md
@@ -118,6 +118,16 @@ Options:
 
           [env: DEFAULT_PROMPT=]
 
+      --dense-path <DENSE_PATH>
+          Optionally, define the path to the Dense module required for some embedding models.
+
+          Some embedding models require an extra `Dense` module which contains a single Linear layer and an activation function. By default, those `Dense` modules are stored under the `2_Dense` directory, but there might be cases where different `Dense` modules are provided, to convert the pooled embeddings into different dimensions, available as `2_Dense_<dims>` e.g. https://huggingface.co/NovaSearch/stella_en_400M_v5.
+
+          Note that this argument is optional, only required to be set if the path to the `Dense` module is other than `2_Dense`. And it also applies when leveraging the `candle` backend.
+
+          [env: DENSE_PATH=]
+          [default: 2_Dense]
+
       --hf-token <HF_TOKEN>
           Your Hugging Face Hub token
 
@@ -191,9 +201,9 @@ Options:
 
           [env: CORS_ALLOW_ORIGIN=]
 
-      -h, --help
+  -h, --help
           Print help (see a summary with '-h')
 
-      -V, --version
+  -V, --version
           Print version
 ```


kozistr:
@alvarobartt Hi! Just for your reference, you may already be aware, Stella v5 model uses `Identity` layer as its activation function for `2_Dense`! [2_Dense/config.json](https://huggingface.co/NovaSearch/stella_en_400M_v5/blob/main/2_Dense/config.json)

backends/candle/src/models/dense.rs
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize)]
+pub struct DenseConfig {
+    in_features: usize,
+    out_features: usize,
+    bias: bool,
+    activation_function: Option<String>,
```
Narsil: Should be `activation_function: DenseActivation` no ? (With a default on DenseActivation,

backends/candle/src/models/dense.rs
```diff
@@ -0,0 +1,78 @@
...
+
+        // Here we cannot leverage HiddenAct, since the activation functions for the
+        // 2_Dense/config.json are defined as PyTorch imports instead, so the deserialization would
+        // be different, as well as the range of commonly used activation functions (mainly tanh
+        // and identity)
+        let activation = match config.activation_function {
+            // e.g. https://huggingface.co/sentence-transformers/LaBSE/blob/main/2_Dense/config.json
+            Some(ref act) if act == "torch.nn.modules.activation.Tanh" => DenseActivation::Tanh,
+            // e.g. https://huggingface.co/NovaSearch/stella_en_400M_v5/blob/main/2_Dense/config.json
+            Some(ref act) if act == "torch.nn.modules.linear.Identity" => DenseActivation::Identity,
```
Narsil: This would be taken care of by serde with `#serde(alias="xxx")]` (or `#[serde(rename="xxx')]` if there is only one variant in the wild.

Narsil (approved):

PR #664, opened 2025-06-27
Add last token pooling support for ORT.

tpendragon:
I really wanted to use a quantized Qwen3-Embedding model, this let me do it.


alvarobartt (dismissed):

backends/ort/src/lib.rs
```diff
@@ -205,7 +205,10 @@ impl Backend for OrtBackend {
                 // CLS pooling
                 Pool::Cls => outputs.slice(s![.., 0, ..]).into_owned().into_dyn(),
                 // Last token pooling is not supported for this model
```
alvarobartt:
```suggestion
```

PR #676, opened 2025-07-04
Support MRL (Matryoshka Representation Learning)

### LINKED ISSUE #673: Feature: Support Matryoshka Representation Learning
Is there any plan to support MRL?  Just like this
```python
from openai import OpenAI
client = OpenAI(api_key="openai-api-key")

client.embeddings.create(
  model="Qwen3-Embedding-8B",
  input="Hello",
  dimensions=1024,
)
```

kozistr:
# What does this PR do?

Fixes #673

Add `dimensions` field to the embed request API spec.

- [x] compatible with OpenAI client
- [x] both http & grpc
- [x] add a test case for the MRL feature
- [x] remove the unused test cases in `router/tests`

```
$ ./target/release/text-embeddings-router --model-id ../Qwen3-Embedding-0.6B/ --pooling last-token --port 8080 --dtype float32 --auto-truncate --max-batch-tokens 512
```

```
from openai import OpenAI

client = OpenAI(api_key='', base_url='http://127.0.0.1:8080')

response = client.embeddings.create(
    model='Qwen3-Embedding-0.6B',
    input='test',
    dimensions=256,
)

len(response.data[0].embedding)
256
```

```
tei_results = np.asarray(
    requests.post(
        'http://127.0.0.1:8080/embed',
        data=json.dumps({'inputs': sentences[0], 'normalize': True, 'dimensions': 256}),
        headers={'Content-type': 'application/json'},
    ).json()
)
tei_results.shape
(1, 256)
```

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@Narsil @alvarobartt


router/src/http/types.rs
```diff
@@ -423,9 +428,15 @@ pub(crate) struct EmbedRequest {
     /// any text to encode.
     #[schema(default = "null", example = "null", nullable = true)]
     pub prompt_name: Option<String>,
+
     #[serde(default = "default_normalize")]
     #[schema(default = "true", example = "true")]
     pub normalize: bool,
+
+    /// The number of dimensions the resulting output embeddings should have. If not set, the orignal
+    /// shape of the representation will be returned.
```
alvarobartt: Maybe a mention here for the Matryoska Representation Learning (MRL)?

alvarobartt (dismissed):
LGTM, thanks for the addition! Also I like the macros for Prometheus metrics, those look cleaner IMO.

P.S. Guessing that the snapshots were removed because those were not used, right?

router/src/http/types.rs
alvarobartt:
```suggestion
    /// The number of dimensions that the output embeddings should have. If not set, the original
    /// shape of the representation will be returned instead.
```

Also there's a typo and I slightly re-worded it (still missing the MRL mention).

kozistr: thanks for it!

kozistr:
> LGTM, thanks for the addition! Also I like the macros for Prometheus metrics, those look cleaner IMO.
> 
> P.S. Guessing that the snapshots were removed because those were not used, right?

oh, sorry for the missing context. you're right! 

looks like [these test cases (router/tests)](https://github.com/huggingface/text-embeddings-inference/tree/main/router/tests) were all commented, so I've removed the snapshots.

updated)
I've just updated the PR description too!

thanks for the pointing this :)

alvarobartt (dismissed):

michaelfeil:
@kozistr QQ: what happens if you set normalize=True + mrl dimensions? Otherwise PR looks good.

kozistr:
> @kozistr QQ: what happens if you set normalize=True + mrl dimensions? Otherwise PR looks good.

@michaelfeil Hi! `MRL` will always run first before the normalization! (+ `normalize` is `True` as a default value)

thanks for double-checking this :)

michaelfeil:
Is there something bringing this to mainline?

core/src/infer.rs
```diff
@@ -283,16 +294,14 @@ impl Infer {
...
-        let histogram = metrics::histogram!("te_embed_inference_duration");
-        histogram.record(response.metadata.inference.as_secs_f64());
+        metrics::counter!("te_embed_success").increment(1);
+        metrics::histogram!("te_embed_duration").record(total_time.as_secs_f64());
+        metrics::histogram!("te_embed_tokenization_duration")
+            .record(response.metadata.tokenization.as_secs_f64());
+        metrics::histogram!("te_embed_queue_duration")
+            .record(response.metadata.queue.as_secs_f64());
+        metrics::histogram!("te_embed_inference_duration")
+            .record(response.metadata.inference.as_secs_f64());
```
Narsil: Can you revert those changes ? They don't seem linked to the PR.

core/src/infer.rs
```diff
@@ -232,20 +229,29 @@ impl Infer {
...
         normalize: bool,
+        dimensions: Option<usize>,
         permit: OwnedSemaphorePermit,
     ) -> Result<PooledEmbeddingsInferResponse, TextEmbeddingsError> {
         let start_time = Instant::now();
 
         if self.is_splade() && normalize {
-            let counter = metrics::counter!("te_request_failure", "err" => "model_type");
-            counter.increment(1);
+            metrics::counter!("te_request_failure", "err" => "model_type").increment(1);
```
Narsil: Sam here.

core/src/infer.rs
```diff
@@ -232,20 +229,29 @@ impl Infer {
...
             let message = "`normalize` is not available for SPLADE models".to_string();
             tracing::error!("{message}");
             return Err(TextEmbeddingsError::Backend(BackendError::Inference(
                 message,
             )));
         }
 
+        if let Some(dimensions) = dimensions {
+            if dimensions == 0 {
+                metrics::counter!("te_request_failure", "err" => "validation").increment(1);
```
Narsil, line 250: Shouldn't they also be smaller than the maximum embedding dimension ?

core/src/infer.rs
kozistr: reverted https://github.com/huggingface/text-embeddings-inference/pull/676/commits/9131071796eacbe04c4546898b0084dbf2fec5e1

core/src/infer.rs
kozistr: reverted https://github.com/huggingface/text-embeddings-inference/pull/676/commits/9131071796eacbe04c4546898b0084dbf2fec5e1

core/src/infer.rs
kozistr: At that time, I was considering silently returning the embedding with the original size when the given dimension is larger than the expected size, like [line 274](https://github.com/kozistr/text-embeddings-inference/blob/feature/mrl/core/src/infer.rs#L274).

On second thought, like you mentioned, it'd be better to raise an error explicitly when the size is larger than expected in terms of validity.

I'll add an extra validation logic to check whether the given size is larger than the size of the embedding. thanks for catching this!

kozistr: added https://github.com/huggingface/text-embeddings-inference/pull/676/commits/a1b1a260a80be807ce3c7fbfca8b465e28c55f9c

Narsil (approved):
LGTM !

core/src/infer.rs
```diff
@@ -262,6 +270,21 @@ impl Infer {
...
+
+                let message =
+                    "`dimensions` should be smaller than the maximum embedding dimension."
+                        .to_string();
+                tracing::error!("{message}");
+
+                return Err(TextEmbeddingsError::Validation(message));
+            }
+
+            response.results.truncate(mrl_dimensions);
```
mkhludnev, line 285: Hi, 
Does it mean that it's up to client to normalize truncated vectors?

mkhludnev: FWIW, it seems it normalizes them by default itself.

alvarobartt: Not really, it does truncate those first then normalizes those as you can see in the code below where if `normalize: true` it will normalize the embeddings; though it's true that it'll normalized over the truncated embeddings rather than on the former embeddings, I'll need to check if that's how it's supposed to work, @kozistr do you know?

Alternatively, to better answer your question, if you set `normalize: true` in the payload to `/embed` it will normalize those, as those are not normalized by default; whereas if you were to use `/v1/embeddings` instead, those would be normalized by default as that's how the OpenAI Embeddings API specification is + there's no `normalize` / `normalization` argument on that spec either.

kozistr: Hi @mkhludnev @alvarobartt!

> it'll normalized over the truncated embeddings rather than on the former embeddings

Yep you're right! it'd be better to normalize on the truncated embeddings for efficiency! So, I just implemented the truncation logic right before the normalization part.

So, the overall flow is like the following,

> original embeddings -> (optional) truncation -> (default: true) l2 normalize

Mathematically, `normalize -> truncate -> normalize` produces the identical result to `truncate -> normalize` anyway, because it only scales magnitude!

PR #678, opened 2025-07-07
spawn prometheus exporter

augustas1:
# What does this PR do?

Spawns Prometheus exporter in the background. Currently metrics are not accessible via `PROMETHEUS_PORT` (default `9000`), only via main port `/metrics` route.

Code addition matches the same what [PrometheusBuilder::install](https://docs.rs/metrics-exporter-prometheus/latest/src/metrics_exporter_prometheus/exporter/builder.rs.html#388) does.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@alvarobartt @Narsil

### DIFF
--- a/router/src/http/server.rs
+++ b/router/src/http/server.rs
@@ -1691,11 +1691,12 @@ pub async fn run(
     });
 
     // See: https://github.com/metrics-rs/metrics/issues/467#issuecomment-2022755151
-    let (recorder, _) = prom_builder
+    let (recorder, exporter) = prom_builder
         .build()
         .context("failed to build prometheus recorder")?;
     let prom_handle = recorder.handle();
     metrics::set_global_recorder(recorder).context("Failed to set global recorder")?;
+    tokio::runtime::Handle::current().spawn(exporter);
 
     // CORS layer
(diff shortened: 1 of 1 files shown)


alvarobartt:
Hey @augustas1, thanks for your contribution! Indeed this seems related to #729, but AFAIK the Prometheus port only applies for gRPC, as it requires to deploy the endpoint over HTTP; whereas on HTTP is not required as we can simply create the `/metrics` route within the default port.

Could you elaborate on why do you think this is required or why? I guess it can be misleading to expose a `--prometheus-port` argument and then not being "used" over HTTP for querying the `/metrics` endpoint but AFAIK the change should be on the argument description or on the docs rather than on the behaviour on whether Prometheus is spawned or not within the HTTP server, thoughts?

Thanks in advance 🤗

augustas1:
Hey @alvarobartt, use case we had - we're using service mesh inside our cluster and mTLS for services. But in order to exclude metrics endpoint from mTLS it has to be on a separate port.

We are starting to use OTLP as well now, are metrics being exported via OTLP as well?

PR #685, opened 2025-07-24
feat: Add an option for specifying model name

vrdn-23:
# What does this PR do?

This PR introduces a new CLI argument, `--served-model-name`, which allows users to specify a custom model name to be returned in responses from the OpenAI-compatible endpoint.

This is particularly useful in scenarios where the model is loaded from a local path (e.g., /data/model) and does not have an inherent name associated with it. By setting `--served-model-name`, users can override the default model identifier (which might be a generic or filesystem-based value) and provide a more descriptive or meaningful name in the API response. This helps improve clarity and consistency, especially when integrating with clients or tools that rely on the model field in the response for tracking or routing purposes.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@Narsil @alvarobartt @kozistr 

I have tested this to the best of my ability, but I'm not sure if I did the gRPC bits correctly, so if someone could help verify that, that would be great!

### DIFF
--- a/router/src/http/server.rs
+++ b/router/src/http/server.rs
@@ -1287,7 +1287,7 @@ async fn openai_embed(
     let response = OpenAICompatResponse {
         object: "list",
         data: embeddings,
-        model: info.model_id.clone(),
+        model: info.served_model_name.clone().unwrap_or_else(|| info.model_id.clone()),
         usage: OpenAICompatUsage {
             prompt_tokens: compute_tokens,
             total_tokens: compute_tokens,
--- a/README.md
+++ b/README.md
@@ -153,6 +153,11 @@ Options:
           Optionally control the number of tokenizer workers used for payload tokenization, validation and truncation. Default to the number of CPU cores on the machine
 
           [env: TOKENIZATION_WORKERS=]
+      
+      --served-model-name <SERVED_MODEL_NAME>
+          The name of the model that is returned when serving OpenAI requests. If not specified, defaults to value in model-id.
+
+          [env: SERVED_MODEL_NAME=]
 
       --dtype <DTYPE>
           The dtype to be forced upon the model
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -1052,6 +1052,7 @@
         "required": [
           "model_id",
           "model_dtype",
+          "served_model_name",
           "model_type",
           "max_concurrent_requests",
           "max_input_length",
@@ -1105,6 +1106,11 @@
           "model_id": {
             "type": "string",
             "description": "Model info",
+            "example": "thenlper/gte-base"
+          },
+          "served_model_name": {
+            "type": "string",
+            "description": "Model name specified by user",
             "example": "thenlper/gte-base"
           },
           "model_sha": {
--- a/docs/source/en/cli_arguments.md
+++ b/docs/source/en/cli_arguments.md
@@ -46,6 +46,11 @@ Options:
 
           [env: DTYPE=]
           [possible values: float16, float32]
+      
+      --served-model-name <SERVED_MODEL_NAME>
+          The name of the model that is returned when serving OpenAI requests. If not specified, defaults to value in model-id.
+
+          [env: SERVED_MODEL_NAME=]
 
       --pooling <POOLING>
           Optionally control the pooling method for embedding models.
--- a/proto/tei.proto
+++ b/proto/tei.proto
@@ -58,6 +58,7 @@ message InfoResponse {
     optional uint32 max_batch_requests = 11;
     uint32 max_client_batch_size = 12;
     uint32 tokenization_workers = 13;
+    optional string served_model_name = 14;
 }
 
 message Metadata {
--- a/router/src/grpc/server.rs
+++ b/router/src/grpc/server.rs
@@ -572,6 +572,7 @@ impl grpc::info_server::Info for TextEmbeddingsService {
             model_id: self.info.model_id.clone(),
             model_sha: self.info.model_sha.clone(),
             model_dtype: self.info.model_dtype.clone(),
+            served_model_name: self.info.served_model_name.clone(),
             model_type: model_type.into(),
             max_concurrent_requests: self.info.max_concurrent_requests as u32,
             max_input_length: self.info.max_input_length as u32,
--- a/router/src/lib.rs
+++ b/router/src/lib.rs
@@ -46,6 +46,7 @@ pub async fn run(
     revision: Option<String>,
     tokenization_workers: Option<usize>,
     dtype: Option<DType>,
+    served_model_name: Option<String>,
     pooling: Option<text_embeddings_backend::Pool>,
     max_concurrent_requests: usize,
     max_batch_tokens: usize,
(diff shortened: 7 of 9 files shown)


vrdn-23:
@alvarobartt Let me know if you have are any changes/feedback for this PR

vrdn-23:
@alvarobartt @Narsil I was wondering if you have any thoughts/concerns/feedback regarding the use-case/implementation for this PR! Let me know if this is something you would like to discuss offline and I am open to that as well!

alvarobartt:
Hey @vrdn-23 thanks for opening this PR and apologies I'm just looking into it now! But I'll check that everything works this week, and happy to support it and add it within the next release 🤗

vrdn-23:
Thanks @alvarobartt for getting back to me. let me know if there are any changes I need to make!

vrdn-23:
@alvarobartt just wanted to check in and see if this was still on your radar for review!

alvarobartt:
Hey @vrdn-23, yes! This is something I'd like to include for Text Embeddings Inference v1.9.0, but I'd like to first make sure that some patches land, apologies for the delay 🙏🏻 

Also given that we add the `--served-model-name`, do you think it makes sense for us to validate that the provided `model` parameter on OpenAI Embeddings API requests i.e., `v1/embeddings`, matches the actual value of either `--model-id` or `--served-model-name` unless provided as empty, similarly to how other providers with OpenAI compatible interfaces do as e.g. vLLM?

vrdn-23:
> Also given that we add the --served-model-name, do you think it makes sense for us to validate that the provided model parameter on OpenAI Embeddings API requests i.e., v1/embeddings, matches the actual value of either --model-id or --served-model-name unless provided as empty, similarly to how other providers with OpenAI compatible interfaces do as e.g. vLLM

I think that would be great! I think by default if the model name isn't specified in vLLM, it still serves the request, so we shouldn't have to change compatibility of the OpenAI format spec. I can add a check for the validation based on the model name specified in `--served-model-name`

vrdn-23:
> matches the actual value of either --model-id or --served-model-name unless provided as empty,

Just to add to this, it does seem vLLM does not check for matching with `--model-id` if `--served-model-name` is specified. So maybe we check for validation like this:

- If `--model-id` is specified and `--served-model-name` is not, validate against `model-id` and return `model-id` in response
- If `--model-id` is specified and `--served-model-name` is specified, validate against `--served-model-name` and return `served-model-name` in response
- If `model` is passed in as empty, return the `--served-model-name` in response else `--model-id` (which is what the PR currently does?)

alvarobartt:
Yes that's right @vrdn-23, I'll validate and approve ✅

vrdn-23:
@alvarobartt sorry this took so long, but I've added the validation we talked about and did some testing! Please do feel free to validate if everything works as expected!

alvarobartt:
Hey @vrdn-23 awesome thanks! I'll release a patch with some fixes this week, then start merging some PRs for v1.9.0 including this one 🤗

router/src/http/server.rs
```diff
@@ -1153,6 +1153,29 @@ async fn openai_embed(
...
     }
 
+    // Validate model name if provided in request
+    if let Some(requested_model) = &req.model {
+        let expected_model = info
+            .served_model_name
+            .as_ref()
+            .unwrap_or(&info.model_id);
+
+        if requested_model != expected_model {
```
michaelfeil: I think this is a breaking Change. Why introduce breaking people prod by saying model not found? 
Adds additional complexity, TEI can only serve one model, and if you're hitting the wrong one you're free to perfom this string assertion (called model name vs response) locally on the caller side.

vrdn-23: Hey @michaelfeil,
Thanks for the review! 

I'm not sure how exactly this becomes a breaking change. The only current scenario where I can see this breaking is if a request had a `model` parameter which was different from the `--model-id` specified (which is probably more of a bug for allowing the request to go through in the first place). The impetus for validation was suggested by @alvarobartt and confirmed here (https://github.com/huggingface/text-embeddings-inference/pull/685#issuecomment-3382514377) 

I would reckon most users probably do not pass in the model parameter at all, cause like you mentioned TEI can only serve one model at a time, but this feature would not alter any behavior related to that, because requests without model specified, should still work even if we specify `--served-model-name` or not. 
 
> if you're hitting the wrong one you're free to perfom this string assertion (called model name vs response) locally on the caller side.

While I realize that this is possible, I think the reason I introduced the PR in the first place, is so that we can have a parameter that returns a proper model-id when running `/v1/embeddings` even if we load it from a local path. For example if you load models from a local path with something like `--model-id /data/model` it would return the model name as `/data/model` in `/v1/embeddings`, which can be confusing when running multiple TEI servers with different models but the same model loading set-up.

Please let me know if this answers your thoughts/concerns. I'd love for @alvarobartt / @kozistr  to also chime in if they have any additional thoughts!

michaelfeil: will break ~100s of models across google, baseten and others.

vrdn-23: > will break ~100s of models across google, baseten and others.

I'm sorry, could you elaborate a bit more? 

The only foreseeable situation where I see this breaking ~100s of models, is if Google/BaseTen is sending a `model` parameter with the request each time and that parameter is not what is specified by `model-id`? Alternatively, this will only get merged with the v1.9.0 release (which will include a lot of bug-fixes and improvements). So folks using v1.8.3 will not have any breaking changes unless they explicitly upgrade. 

The alternative solution is to avoid doing the validation unless `--served-model-name` is explicitly set, but this is not the standard adopted by other serving frameworks like vLLM. I'll let @alvarobartt chime in from here on.

kozistr: Hi! IMHO, since TEI only serves a single model, and the `model` parameter is unlikely to be used frequently, I'm a bit worried that having strict validation logic might be unnecessary and trip up users who do send it.

Maybe we could just log a warning for now instead?

Happy to hear your thoughts or comments!

alvarobartt: Yes, nice points! Then let's remove the validation on the provided `model` parameter and let's just allow anything as I/O and then set the `model` in the output to whatever value is set on either `--served-model-name` or `--model-id` in that order of priority.

I'm happy to take on this PR if you are short on time @vrdn-23, thanks a lot for your contributions! 🤗

vrdn-23: That would be awesome if you would like to take this up @alvarobartt. I won’t be able to get to this until Tuesday 😅

alvarobartt: Sure it's fine, I'll take it from here, thanks a lot for the contribution @vrdn-23 (and the patience) 🙏🏻

alvarobartt: @vrdn-23 I did some changes on top, but I'd need you to grant me (@alvarobartt) write permission to your fork as otherwise I cannot push the changes. Alternatively I can bring those into a new branch and commit there.

vrdn-23: Just added you as a collaborator! Thanks for helping out @alvarobartt !

PR #689, opened 2025-07-29
adjust HPU warmup: use dummy inputs with shape more close to real scenario 

kaixuanliu:
In original implementation, we use dummy inputs with shapes like [1,128], [1,256],[2,128],[2,256] to do warmup, aiming to generate recipe cache in warmup stage. And in real serving scenario, we padding the input_ids/attention_masks to shapes we cached in warmup stage. However, we found precision issue for reranker models following [tei docs](https://github.com/huggingface/text-embeddings-inference?tab=readme-ov-file#using-re-rankers-models) . We think it may be because we used wrong graphs during replay stage. Hence we adjust the `create_warmup_batch` function in this PR, to make the dummy inputs more close to real scenario, hence during warmup stage, these inputs will also be padded in python backend and will generate right recipe caching/graph, which will be the same with serving stage. We made several round experiments w/ this PR, and the wrong output issue disappears after this PR.

### DIFF
--- a/backends/src/lib.rs
+++ b/backends/src/lib.rs
@@ -168,7 +168,7 @@ impl Backend {
             }
         }
         for shape in shapes.iter() {
-            let batch = self.create_warmup_batch(*shape, max_token as u32);
+            let batch = self.create_warmup_batch(*shape, max_token as u32, seq_bucket_size as u32);
             match &self.model_type {
                 ModelType::Classifier => self.predict(batch).await.map(|_| ()),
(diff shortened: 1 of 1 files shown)


kaixuanliu:
@regisss , pls help review, thx!

regisss (approved):
LGTM

regisss:
cc @Narsil 
The methods `warmup_hpu` and `create_warmup_batch` are only used on HPU

Issue #704, opened 2025-08-20
Qwen/Qwen3-Embedding-0.6B output error


asdfgh19:
### System Info

I run Qwen/Qwen3-Embedding-0.6B using the following command：
`docker run -d --name text-embeddings-inference_server -p 8089:80 -v "$VOLUME":/data -e HF_ENDPOINT=https://hf-mirror.com --pull always ghcr.io/huggingface/text-embeddings-inference:cpu-1.7.4 --revision refs/pr/27 --pooling last-token --max-batch-tokens 512 --model-id Qwen/Qwen3-Embedding-0.6B`

The following request returns different results when called multiple times, sometimes returning result 1 and sometimes returning result 2.

`
curl --location 'http://localhost:8089/embed' --header 'Content-Type: application/json' --data '{
    "inputs": [
        "{\"name\": \"风场\", \"type\": \"model\", \"id\": \"EnOS_Wind_Farm\", \"desc\": \"风场模型\"}",
        "{\"name\": \"发电量\", \"type\": \"metric\", \"id\": \"ActiveProduction\", \"model\": \"EnOS_Wind_Farm\", \"model_name\": \"风场\", \"data_type\": \"double\", \"unit\": \"s\"}",
        "{\"name\": \"发电量\", \"type\": \"metric\", \"id\": \"ActiveProduction\", \"model\": \"EnOS_Solar_Site\", \"model_name\": \"光伏场站\", \"data_type\": \"double\", \"unit\": \"s\"}"
    ]
}'
`

result 1:
`[[-0.03803645, 0.010309944, ...], [...], [...]]`

result 2:
`[[-0.02697843, -0.005487643, ...], [...], [...]]`

### Information

- [x] Docker
- [ ] The CLI directly

### Tasks

- [x] An officially supported command
- [ ] My own modifications

### Reproduction

1. docker run
2. call embed api

### Expected behavior

Multiple requests with the same input return the same result

asdfgh19:
In addition, concurrent calls from multiple threads will also affect each other, resulting in incorrect results

alvarobartt:
Hey @asdfgh19 thanks for flagging! I'm already looking into it, but TL;DR the padding side for `OrtBackend` defaults to `right`, whereas the one used for Qwen3-Embedding models is `left`; meaning that when padding multiple sequences since the padding side is wrong, which shouldn't affect the pooling but might as well be the case for those diffs.

[...]

PR #705, opened 2025-08-21
Add `padding_side` handling in `OrtBackend`

alvarobartt:
# What does this PR do?

This PR adds the `padding_side` from the `tokenizer_config.json` if applicable, otherwise it defaults to `padding_side: "right"` to handle the scenarios where the `padding_side` is other than "right", as e.g. https://huggingface.co/onnx-community/Qwen3-Embedding-0.6B-ONNX to ensure parity with the inputs and the outputs.

This PR also updates the input preparation and pooling strategies accordingly, so that those are applied one way or another based on the padding side, given that the pooling should be padding-agnostic, but with the padding side information we can efficiently apply the pooling strategies instead.

Additionally, this PR fixes the last-token pooling for the `OrtBackend` which was leading to issues (unrelated to the `padding_side`) as e.g. https://github.com/huggingface/text-embeddings-inference/issues/704

As some other, smaller but still relevant changes, this PR:
- Adds `OrtBackend::prepare_inputs` to prepare the `ndarray`s for the `input_ids`, `attention_mask`, etc. within a function to be reused for both `OrtBackend::embed` and `OrtBackend::predict` to prevent from duplicating the code
- Adds `OrtBackend::prepare_ort_inputs` to go from `ndarray`s to `ort::inputs!`, and the reason is the same as per the function above.
- Adds `ModelInputs` to capture all the inputs within the same struct so that it can be easily managed
- Adds `Config` to handle also the `pad_token_id`, because it was previously set as 0 whilst that was not true for all the models, so now the `pad_token_id` is read from the `config.json` if there, otherwise it falls back to `eos_token_id`, or alternatively sets it to 0.

## Before submitting

## Who can review?

@Narsil

### DIFF
--- a/backends/ort/src/lib.rs
+++ b/backends/ort/src/lib.rs
@@ -36,6 +36,14 @@ impl std::str::FromStr for PaddingSide {
     }
 }
 
+struct ModelInputs {
+    pub input_ids: ndarray::Array2<i64>,
+    pub attention_mask: ndarray::Array2<i64>,
+    pub token_type_ids: Option<ndarray::Array2<i64>>,
+    pub position_ids: Option<ndarray::Array2<i64>>,
+    pub input_lengths: ndarray::Array1<f32>,
+}
+
 pub struct OrtBackend {
     session: Mutex<Session>,
 
@@ -175,19 +183,7 @@ impl OrtBackend {
         &self,
         batch: &Batch,
         padding_side: &PaddingSide,
-    ) -> Result<
-        (
-            (
-                ndarray::Array2<i64>,
-                ndarray::Array2<i64>,
-                Option<ndarray::Array2<i64>>,
-                Option<ndarray::Array2<i64>>,
-                ndarray::Array1<f32>,
-            ),
-            bool,
-        ),
-        BackendError,
-    > {
+    ) -> Result<(ModelInputs, bool), BackendError> {
         let batch_size = batch.len();
         let max_length = batch.max_length as usize;
         let elems = batch_size * max_length;
@@ -313,13 +309,13 @@ impl OrtBackend {
         let input_lengths = ndarray::Array1::from_vec(input_lengths);
 
         Ok((
-            (
+            ModelInputs {
                 input_ids,
                 attention_mask,
                 token_type_ids,
                 position_ids,
                 input_lengths,
-            ),
+            },
             masking,
         ))
     }
@@ -337,23 +333,17 @@ impl OrtBackend {
             "attention_mask" => ort::value::Tensor::from_array(attention_mask).e()?,
         ];
 
-        match token_type_ids {
-            Some(token_type_ids) => {
-                let token_type_ids = ort::value::Tensor::from_array(token_type_ids).e()?;
-                inputs.push((
-                    self.token_type_ids_key.clone().into(),
-                    token_type_ids.into(),
-                ));
-            }
-            None => (),
+        if let Some(token_type_ids) = token_type_ids {
+            let token_type_ids = ort::value::Tensor::from_array(token_type_ids).e()?;
+            inputs.push((
+                self.token_type_ids_key.clone().into(),
+                token_type_ids.into(),
+            ));
         }
 
-        match position_ids {
-            Some(position_ids) => {
-                let position_ids = ort::value::Tensor::from_array(position_ids).e()?;
-                inputs.push(("position_ids".into(), position_ids.into()));
-            }
-            None => (),
+        if let Some(position_ids) = position_ids {
+            let position_ids = ort::value::Tensor::from_array(position_ids).e()?;
+            inputs.push(("position_ids".into(), position_ids.into()));
         }
 
         if self.past_key_values {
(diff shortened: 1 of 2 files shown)


backends/ort/src/lib.rs
```diff
@@ -11,23 +12,56 @@ use text_embeddings_backend_core::{
...
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.to_lowercase().as_str() {
+            "left" => Ok(PaddingSide::Left),
+            "right" => Ok(PaddingSide::Right),
+            _ => Err(format!("unrecognized `padding_side` value: {}", s)),
+        }
+    }
+}
```
kozistr: maybe we could utilize `serde` like this!

```
#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(rename_all = "lowercase")]
enum PaddingSide {
    Left,
    Right,
}
```

alvarobartt: Yes I know, the thing here is that we're using `parse` for simplicity (as `.parse::<PaddingSide>()`), and that won't work indeed... I might try to find a more serde-like solution before marking the PR as ready to review 🤗

backends/ort/src/lib.rs
```diff
@@ -11,23 +12,45 @@ use text_embeddings_backend_core::{
...
-    pub num_hidden_layers: usize,
-    pub num_key_value_heads: usize,
+pub struct Config {
+    pub pad_token_id: Option<usize>,
+    pub eos_token_id: Option<usize>,
+    // NOTE: the fields below are only required when the ONNX model expects the `past_key_values`
+    // as input i.e., whenever the ONNX model has been exported with optimized MHA nodes
+    pub hidden_size: Option<usize>,
+    pub num_hidden_layers: Option<usize>,
+    pub num_key_value_heads: Option<usize>,
```
Narsil: Don't put them as option, put them as real values (maybe keep options on fields which are regularly missing and are not import).

Then you can change your code to be simply parsing the Config, and raise a proper error when it fails (where you can inspect the raw Value thing if you want to handle totally the error message, the the serde_json error is usually quite clear, you only need to add information about which file doesn't have the necessary config.

backends/ort/src/lib.rs
```diff
@@ -94,175 +117,310 @@ impl OrtBackend {
...
+                serde_json::from_str::<serde_json::Value>(&content)
+                    .ok()?
+                    .get("padding_side")
+                    .and_then(|v| serde_json::from_value::<PaddingSide>(v.clone()).ok())
+            })
+            .flatten()
+            .unwrap_or_else(|| {
+                tracing::warn!("Could not determine `padding_side` from `tokenizer_config.json`, hence using `right` padding by default.");
+                PaddingSide::Right
+            });
```
Narsil: Seems like this could be simplified quite a bit by expecting things to exist, and only yelling on error, shoudl remove a lot of indentation.

backends/ort/src/lib.rs
```diff
@@ -94,175 +117,310 @@ impl OrtBackend {
             }
         }
 
-        let past_key_values_config = match past_key_values {
-            true => {
-                let path = model_path.join("config.json");
-                if !path.exists() {
+        let config_path = model_path.join("config.json");
```
Narsil: Is it really ever OK to be missing a config.json ?

backends/ort/src/lib.rs
```diff
@@ -303,33 +461,107 @@ impl Backend for OrtBackend {
...
+                            .unwrap()
+                            .into_dyn()
+                        } else {
+                            outputs.slice(s![.., 0, ..]).into_owned().into_dyn()
+                        }
+                    }
+                    PaddingSide::Right => outputs.slice(s![.., 0, ..]).into_owned().into_dyn(),
+                },
+                Pool::LastToken => match self.padding_side {
+                    // NOTE: when using left-padding, the last-token is always in the last position
```
Narsil: Feels like there's a lot of switching on padding side. I haven't carefully looked at each line, but it seems to me that the code could be made significantly simpler by simply allocating things and using a different offset of insertion (overwriting) using the padding side.

Something like

```rust
let offset = if padding_side == Side::Left {0} else {max_length - length};
for (i, item) in  elements.iter().enumerate(){
     input_ids[i + offset] = item;
}
```

Narsil (commented):
Looks alright, but I think we can simplify further.

backends/ort/src/lib.rs
alvarobartt: The thing is that if the `config.json` doesn't exist the error is raised by the `text-embeddings-router` when downloading the artifacts, so here it should always be there yes

backends/ort/src/lib.rs
alvarobartt: Hmm yes it could be the case for some, the issue is that given the padding_side, we can apply the pooling in a most performant approach as in e.g. last-token pooling when left-padding it's literally the last token in the sequence, but when padding is right we need to iterate over each sequence to identify where it ends and then capture what's the last-token accordingly; this is why I kept one implementation per padding_side, but happy to unify those into a single match even if either right or left might be slightly less performance as in requiring more ops to obtain, which I'm also fine with, as that'd simplify the code a bit.

backends/ort/src/lib.rs
alvarobartt: Ok I went through those and apparently only `hidden_size` and `num_hidden_layers` don't need to be optional, the rest are indeed optional, so we have different alternatives:

1. Set the types as those should i.e. handling optionals when it's the case
2. Add an intermediate validation schema that validates given the constraints:
    - `pad_token_id` if provided, else fallback to `eos_token_id` if provided, else fallback to 0
    - `num_key_value_heads` if provided, else fallback to `num_hidden_layers`

This being said, 2 might be nicer as we validate and assign with an intermediate schema, then we ensure that what's on `Config` can be access safely as e.g. `config.pad_token_id` without having to handle the `None` case; but also adds more lines of code and leverages more `serde` "magic" which might not be great in the long term. As per 1, we can go that path if preferred, but we might need to propagate the fields in `OrtBackend` rather than via `Config`.

alvarobartt: Yes, I've just changed `pad_token_id`, `eos_token_id` and `num_key_value_heads` to `Option<usize>` as those don't need to necessarily be present, and handle those unwraps were applicable, whilst providing the `Config` to Self instead, so that it can be accessed from it rather than adding extra fields as e.g. `pad_token_id`

backends/ort/src/lib.rs
Narsil: Fair enough.

PR #742, opened 2025-10-22
Feat: Add neuron backend to TEI

JingyaHuang:
# What does this PR do?

This PR adds Neuron as a device option for the TEI Python backend.

* Both pre-compiled ckpt and compilation on the fly are supported


backends/src/lib.rs
```diff
@@ -68,6 +68,13 @@ fn is_hpu() -> bool {
     }
 }
 
+fn is_neuron() -> bool {
+    match Command::new("neuron-ls").output() {
+        Ok(output) => output.status.success(),
+        Err(_) => false,
+    }
+}
+
```
alvarobartt: > The following is subject to leaving the `python` feature only for `python-habana` and creating a `python-neuron` feature instead; so that `python` feature is deprecated in favor of `python-habana` in the next releases, cc @kaixuanliu

I don't dislike this, but I'd prefer to remove it and leave it to the backend to fail i.e., if user compiles with `--features python-neuron` and either the `model.neuron` file is not there, or the device is not Inferentia 2 (or any other AWS Neuron supported device), the initialization of the backend will fail. But I'll leave this snippet here to only perform the download of the files required for the backend, rather than on backend validation, thoughts? @JingyaHuang

backends/src/lib.rs
```diff
@@ -736,6 +775,20 @@ async fn download_onnx(api: Arc<ApiRepo>) -> Result<Vec<PathBuf>, ApiError> {
     }
 }
 
+async fn download_neuron(api: &ApiRepo) -> Result<Vec<PathBuf>, ApiError> {
+    let mut model_files: Vec<PathBuf> = Vec::new();
+
+    tracing::info!("Downloading `model.neuron`");
+    match api.get("model.neuron").await {
```
alvarobartt, line 777: Not sure how common this is, but I'd extend support for sharded AWS Neuron model files.

backends/src/lib.rs
```diff
@@ -418,16 +425,48 @@ async fn init_backend(
     if let Some(api_repo) = api_repo.as_ref() {
         if cfg!(feature = "python") || cfg!(feature = "candle") {
```
alvarobartt: TL;DR Add an if-statement either above or below this one to capture `if cfg!(feature = "python-neuron")`, and inside attempt to download the `model.neuron` file first, if not there then warn about runtime compilation and download the `model.safetensors`.

backends/src/lib.rs
```diff
@@ -418,16 +425,48 @@ async fn init_backend(
...
+                        https://huggingface.co/docs/optimum-neuron/en/model_doc/sentence_transformers/overview "
+                    );
+                    // Fall back to downloading regular model files for on-the-fly compilation
+                    if download_safetensors(api_repo.clone()).await.is_err() {
+                        tracing::warn!(
+                            "safetensors weights not found. Using `pytorch_model.bin` instead."
+                        );
+                        tracing::info!("Downloading `pytorch_model.bin`");
+                        api_repo
+                            .get("pytorch_model.bin")
```
alvarobartt, line 442: Will both `pytorch_model.bin` and `model.safetensors` be compiled with AWS Neuron the same way, or is there any implication? I guess not, but asking just in case.

backends/python/server/text_embeddings_server/models/neuron/__init__.py
alvarobartt, line 1: Maybe there's a "better" way of splitting both Intel Habana and AWS Neuron on e.g. different subdirectories under `text_embeddings_server`?

Dockerfile-neuron
```diff
@@ -0,0 +1,190 @@
...
+
+
+FROM neuron AS grpc
+
+COPY --from=grpc-builder /usr/src/target/release/text-embeddings-router /usr/local/bin/text-embeddings-router
+
+ENTRYPOINT ["text-embeddings-router"]
+CMD ["--json-output"]
+
+FROM neuron
```
alvarobartt: Nit

```suggestion
FROM neuron AS http
```

Dockerfile-neuron
```diff
@@ -0,0 +1,190 @@
...
+    controlnet-aux \
+    huggingface_hub==${HUGGINGFACE_HUB_VERSION} \
+    hf_transfer \
+    datasets==${DATASETS_VERSION} \
+    optimum-neuron==${OPTIMUM_NEURON_VERSION} \
+    sentence_transformers==${SENTENCE_TRANSFORMERS} \
+    peft==${PEFT_VERSION} \
+ && rm -rf ~/.cache/pip/*
+
+
```
alvarobartt: Nit

```suggestion

```

Dockerfile-neuron
```diff
@@ -0,0 +1,190 @@
...
+    diffusers==${DIFFUSERS_VERSION} \
+    compel \
+    controlnet-aux \
+    huggingface_hub==${HUGGINGFACE_HUB_VERSION} \
+    hf_transfer \
+    datasets==${DATASETS_VERSION} \
+    optimum-neuron==${OPTIMUM_NEURON_VERSION} \
+    sentence_transformers==${SENTENCE_TRANSFORMERS} \
+    peft==${PEFT_VERSION} \
+ && rm -rf ~/.cache/pip/*
```
alvarobartt, line 175: Not something to tackle in this PR maybe, but I'd rather rely on a `lock` file here instead of those, so it might be worth consider re-opening https://github.com/huggingface/text-embeddings-inference/pull/587?

cc @regisss and @kaixuanliu as this was something mentioned in the past, but apparently it was failing on Intel HPUs (?)

Dockerfile-neuron
```diff
@@ -0,0 +1,190 @@
+FROM lukemathwalker/cargo-chef:latest-rust-1.85-bookworm AS chef
```
alvarobartt: This has been recently updated to:

```suggestion
FROM lukemathwalker/cargo-chef:latest-rust-1.92-bookworm AS chef
```

alvarobartt (commented):
Thanks a lot @JingyaHuang, great work! 🤗

Dockerfile-neuron
regisss: I think it should work on HPU, not sure why it failed at that time. so don't hesitate to go that way, and if you have a lock file you would like me to test on HPU, happy to do it :)

backends/python/server/text_embeddings_server/models/neuron/__init__.py
JingyaHuang: Yeah, I would put neuron related modeling code under a folder named `backends/python/server/text_embeddings_server/models/neuron`, but for habana, I'm not very confident which ones are habana-only.

backends/src/lib.rs
JingyaHuang: it's nice if we can create a feature neuron-only, will add it.

backends/src/lib.rs
JingyaHuang: Yes, they will be handled the same way by optimum-neuron.

Dockerfile-neuron
alvarobartt: Thanks @regisss, I'll restart Nico's PR to add `uv` support instead, and ping you when done for testing 🤗

kandakji:
Any update on neuron support?

JingyaHuang:
Hi @kandakji, this PR is on hold pending upcoming changes in the Neuron SDK. We're currently in a transition phase, and some code in the PR is likely to shift. I'll probbably open another PR, will ping you there.

PR #753, opened 2025-11-06
Default `HiddenAct::Gelu` to GeLU + tanh in favour of GeLU erf 

vrdn-23:
# What does this PR do?

This PR unifies `HiddenAct::Gelu` to use GeLU + tanh over exact GeLU (aka. GeLU erf), despite Transformers deserializing `hidden_act` from `gelu` as GeLU erf and `gelu_pytorch_tanh` (and others, see snippet below); all the `gelu` variants will deserialize to GeLU + tanh.

```python
ACT2CLS = {
    "gelu": GELUActivation,
    "gelu_10": (ClippedGELUActivation, {"min": -10, "max": 10}),
    "gelu_fast": FastGELUActivation,
    "gelu_new": NewGELUActivation,
    "gelu_python": (GELUActivation, {"use_gelu_python": True}),
    "gelu_pytorch_tanh": GELUTanh,
    "gelu_python_tanh": (GELUTanh, {"use_gelu_tanh_python": True}),
    "gelu_accurate": AccurateGELUActivation,
    ...
```

This might lead to subtle differences on the produced logits, it won't impact the performance or inference quality (see the modified snapshots). Additionally, adding GeLU erf will be problematic as it will lead to differences when running the same model on CPU or Metal, and CUDA; given that cuBLASLt will only implement GeLU and ReLU.

## Before submitting

- [x] Was this discussed/approved via a GitHub issue or the [forum](https://discuss.huggingface.co/)? Please add a link to it if that's the case.

## Who can review?

@Narsil OR @alvarobartt OR @kozistr

### DIFF
--- a/backends/candle/tests/snapshots/test_flash_gte__gte_classification_single.snap
+++ b/backends/candle/tests/snapshots/test_flash_gte__gte_classification_single.snap
@@ -2,4 +2,4 @@
 source: backends/candle/tests/test_flash_gte.rs
 expression: predictions_single
 ---
-- - -0.7363281
+- - -0.7426758


vrdn-23:
Okay so after some more digging, it seems one of the main reasons to not change this would be speed of `gelu_erf()` compared to `gelu()`. I was digging through the candle repository and saw some relevant issues here. I can run some benchmarks later this week to see if there is still a performance loss. 

https://github.com/huggingface/candle/issues/1062
https://github.com/huggingface/candle/issues/2418
https://github.com/huggingface/candle/issues/1926#issuecomment-2016601177

vrdn-23:
I was able to address the issue with latency by raising this PR. https://github.com/huggingface/candle/pull/3168
So hopefully we can make this change without any loss in performance now!

kozistr:
Wow awesome work https://github.com/huggingface/candle/pull/3168 @vrdn-23! 

I thought the compiler uses constant propagation, so `sqrt(1/2)` part doesn't actually execute at every call.

I roughly assume that TEI uses the approximate gelu (~= new gelu) for faster inference, while there's a marginal difference between the variants.

We could benchmark the speed by updating the `candle-nn` version to the latest commit, which includes your `candle-nn` PR, [here](https://github.com/huggingface/text-embeddings-inference/blob/main/Cargo.toml#L60C69-L60C109)!

One minor point for discussion is that, if the latency stays comparable, it might make sense to keep this implementation.

Anyway, looks great to me!

I’d appreciate any feedback or thoughts you have!

vrdn-23:
Sorry for the late response, but I was away for Thanksgiving break @kozistr !

> I roughly assume that TEI uses the approximate gelu (~= new gelu) for faster inference, while there's a marginal difference between the variants.

I think that might have been true previously, but I think now that https://github.com/huggingface/candle/pull/3168 has been merged, the gelu_erf (old gelu) implementation is actually faster than the new one. It would also be functionally most similar in producing outputs with the existing models, so since it seems to be a win in terms of both quality and latency, I would argue that maybe we stick to the consistent implementation across frameworks.

Would love to hear your thoughts @alvarobartt @Narsil

vrdn-23:
Just wanted to add a note that #784 should probably be merged in before this (if accepted), so that the speed-up gained by https://github.com/huggingface/candle/pull/3168 can be utilized

alvarobartt (dismissed):

vrdn-23:
Thanks for the update and picking up my slack @alvarobartt. My day job has been hectic for the past month, so it's been hard to pick up the work on these PRs. Appreciate you helping with these! <3

michaelfeil:
Background: I am pretty sure that cublasLT cannot fuse any kind of actication funcitons. 

The difference in using gelu, fast_gelu is low. 
This matters at most in training. Candle has a fp16 (no bf16! support). 

Can you provide a trained (non random init model with in-distribution activations), that has a embedding of cosine similarity <=0.99. I assume using GELU. In commercial inference products I used, I reverted the GELU_exact -> GELU_fast for this exact reason. 

I'd be skeptical that this not a perf regression, but am otherwise not biased.

alvarobartt:
> The difference in using gelu, fast_gelu is low. This matters at most in training. Candle has a fp16 (no bf16! support).

Indeed I'm working on adding BF16 support as we speak, started with a draft last week in order to support https://huggingface.co/voyageai/voyage-4-nano (and the rest of their recently released embedding models).

> I'd be skeptical that this not a perf regression, but am otherwise not biased.

I'm also not 100% sure yet on adding this change over, skipping `gelu_erf` and just aliasing every `gelu` variant in the `config.json` to `gelu()` instead; meaning that we simply skip the `HiddenAct::GeluErf` variant in favour of `HiddenAct::Gelu`. Also taking into consideration that `gelu` is faster (in theory) than `gelu_erf`, so I might revert the `GeluErf` variant in favour of `Gelu`, thoughts @michaelfeil?

vrdn-23:
The reason I made this PR was so because I could get exact numbers for reproducibility with DebertaV2SequenceClassification models and have TEI match the same as Transformers. I am not sure if there is a performance regression by running this on CUDA, but atleast on metal, `gelu_erf` should be faster than `gelu` now. I also think that if we are making everything a single Activation with `Gelu` we should leave a substantial comment on why that decision was made and also maybe have some benchmarks for what could be the min/max/median/avg difference in embeddings produced by models from using `gelu` instead of `gelu_erf` compared to transformers.

michaelfeil:
https://docs.rs/cudarc/latest/x86_64-pc-windows-msvc/cudarc/cublaslt/safe/enum.Activation.html Supports Gelu and Relu. Other models will run 10% slower. Usually, differences will be introduced by other modeling issues, not activation. 

As said, they will affect snapshot tests, but gelu_exact or gelu will produce the same scores etc on MTEB and other scores. Deberta was recently added, and multiple other reasons could be the cause.

michaelfeil:
I think the reason here might be that using non-Gelu/Relu, the activation in CublasLT can be dropped? Not sure what happens with these activations, since we just pass it inside cublas lt.

PR #767, opened 2025-11-26
Serialization in tokio thread instead of blocking thread, 50% reduction in latency for small models

michaelfeil:
# What does this PR do?

There is a performance bug introduced first by the initial release of TEI. The embeddings are calculated in a std::thread, with the intend to not block the main backend. While this is a fine idea, it would be much better to do it in a threadPool. tokio_spawn_blocking is such a thread pool with lazy warmup, therefore its a good idea to just use the runtimes pool for that. 

Running a small model yields around ~20% more performance, some cases also 50%. Also leads to 15% thoughput improvments for small models. 
```
text-embeddings-router --model-id TaylorAI/bge-micro --max-batch-tokens 280960 --port 7998 --max-client-batch-size 512
```

mainline
```
1 token requests, 512 clients:
Requests per second:    709.84 [#/sec] (mean)
512 token requests, 32 clients
Requests per second:    93.99 [#/sec] (mean)
1 token request, 1 client:
Time per request:       1.865 [ms] (mean)
```

This branch
```
1 token requests, 512 clients:
Requests per second:    888.68 [#/sec] (mean)
512 token requests, 32 clients
Requests per second:    118.40 [#/sec] (mean)
1 token request, 1 client:
Time per request:       1.267 [ms] (mean, across all concurrent requests)
```

Fixes # (issue)

## Before submitting

## Who can review?

### DIFF
--- a/core/src/infer.rs
+++ b/core/src/infer.rs
@@ -550,8 +550,8 @@ async fn backend_task(backend: Backend, mut embed_receiver: mpsc::Receiver<NextB
             ModelType::Classifier => {
                 let results = backend.predict(batch.1).await;
 
-                // Handle sending responses in another thread to avoid starving the backend
-                std::thread::spawn(move || match results {
+                // Handle sending responses in a blocking task to avoid starving the backend
+                tokio::task::spawn_blocking(move || match results {
                     Ok((mut predictions, inference_duration)) => {
                         batch.0.into_iter().enumerate().for_each(|(i, m)| {
                             let infer_metadata = InferMetadata {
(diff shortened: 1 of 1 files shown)


michaelfeil:
mosty came up with the idea to look into this when finding: https://github.com/huggingface/text-embeddings-inference/pull/766 . However, in  https://github.com/huggingface/text-embeddings-inference/pull/766, its actually a good idea to just use std::thread, since the process is running forever. The treads in the PR are short-lived, so better use the tokio pool.

michaelfeil:
openai codex review: https://github.com/michaelfeil/text-embeddings-inference/pull/1#issuecomment-3579866298

PR #777, opened 2025-12-17
feat: add varlen attention on cpu

michaelfeil:
# What does this PR do?

This PR brings varlen-flash-attention to CPU/Metal. (Its not softmax-fused / Flash), but at least its not-padded, so we don't do OOM. 

Fixes # (issue)

## Before submitting

## Who can review?

### DIFF
--- a/backends/candle/src/flash_attn_cpu.rs
+++ b/backends/candle/src/flash_attn_cpu.rs
@@ -4,6 +4,9 @@ use candle::Tensor;
 /// This implements standard attention computation for CPU and supports all model types
 /// actually not "flash" fused, but using no-padding will lower memory usage and flops on CPU
 #[allow(dead_code)]
+#[allow(clippy::too_many_arguments)]
+#[allow(clippy::type_complexity)]
+#[allow(clippy::needless_range_loop)]
 pub fn flash_attn_varlen_cpu(
     q: &Tensor,
     k: &Tensor,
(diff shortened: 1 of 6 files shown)


backends/candle/src/flash_attn_cpu.rs
```diff
@@ -0,0 +1,2539 @@
...
+
+        // Apply softmax scale
+        let scale_tensor = Tensor::new(softmax_scale as f32, device)?;
+        let mut attention_scores =
+            attention_scores.mul(&scale_tensor.broadcast_as(attention_scores.shape())?)?; // [num_heads, seq_len_q, seq_len_k]
+
+        // Apply causal mask if requested
+        if causal {
+            let causal_mask = create_causal_mask_batch(seq_len_q, seq_len_k, num_heads, device)?;
+            attention_scores = attention_scores.add(&causal_mask)?;
```
kozistr, line 115: i was just wondering that it looks like `causual_mask` and `window_mask` below are always `fp32` type while `attention_scores` could be fp16. I'm not sure if I'm right, it might fail due to a type mismatch!

kozistr (commented):
aside from one thing, everything else looks good to me!

backends/candle/src/flash_attn_cpu.rs
michaelfeil: Thanks, you are right. i figured that out too.

michaelfeil: In this case its always fp32, but for apple metal backend it could be fp16 afaik.

michaelfeil:
I am going to mark this PR as draft. I implemented a pretty fast attention primitive here: https://github.com/huggingface/candle/pull/3250 once that is merged (which i am eagerly waiting for) we can do a simple copy of the function here (without the tests).

PR #778, opened 2025-12-18
simplication: flash feature config. 

michaelfeil:
makes flash-attn-3 and flash-attn-cpu easier to add.

Does not change any logic.

### DIFF
--- a/backends/candle/src/lib.rs
+++ b/backends/candle/src/lib.rs
@@ -118,12 +118,12 @@ enum Config {
 /// Helper function to check if flash attention should be used
 /// require_fa2_capabilities: true for models that need flash attention v2, false for v1/v2 compatible models
 fn should_use_flash_attention(require_fa2_capabilities: bool) -> bool {
-    let flash_attn_enabled = &std::env::var("USE_FLASH_ATTENTION").unwrap_or("true".to_string()).to_lowercase() == "true";
-    
     if cfg!(not(feature = "cuda")) {
         // if not cuda support, always false for now.
         return false;
     };
+
+    let flash_attn_enabled = &std::env::var("USE_FLASH_ATTENTION").unwrap_or("true".to_string()).to_lowercase() == "true";
     // cuda
     if require_fa2_capabilities {
         cfg!(feature = "flash-attn") && flash_attn_enabled


backends/candle/src/lib.rs
```diff
@@ -115,6 +115,23 @@ enum Config {
...
 
+/// Helper function to check if flash attention should be used
+/// require_fa2_capabilities: true for models that need flash attention v2, false for v1/v2 compatible models
+fn should_use_flash_attention(require_fa2_capabilities: bool) -> bool {
+    let flash_attn_enabled = &std::env::var("USE_FLASH_ATTENTION").unwrap_or("true".to_string()).to_lowercase() == "true";
+    
+    if cfg!(not(feature = "cuda")) {
+        // if not cuda support, always false for now.
+        return false;
+    };
```
kozistr: It's a nit; how about moving line 121 right after line 126 to the early-return non-cuda case?

```suggestion    
    if cfg!(not(feature = "cuda")) {
        // if not cuda support, always false for now.
        return false;
    };

    let flash_attn_enabled = &std::env::var("USE_FLASH_ATTENTION").unwrap_or("true".to_string()).to_lowercase() == "true";
```

backends/candle/src/lib.rs
```diff
@@ -115,6 +115,23 @@ enum Config {
     XlmRoberta(BertConfig),
 }
 
+/// Helper function to check if flash attention should be used
+/// require_fa2_capabilities: true for models that need flash attention v2, false for v1/v2 compatible models
+fn should_use_flash_attention(require_fa2_capabilities: bool) -> bool {
```
alvarobartt, line 120: IMO `require_fa2_capabilities` here might be a bit misleading when reading through the code and seeing `should_use_flash_attention(false)` (which means it should but Flash Attention v1).

I did a similar thing in #809 at https://github.com/huggingface/text-embeddings-inference/blob/c19087e37af22374a58c56ea03ca5ef30b09da34/backends/candle/src/flash_attn.rs#L23-L40, so something along those lines might feel a bit more intuitive? WDYT @michaelfeil and @kozistr?

Otherwise as I don't have a strong opinion either I'm happy to merge as-is, my only concerns are readability + if we add other Flash Attention features we might need to change the function signature over just adding a function (?)

kozistr: I think that if we end up with multiple FA versions, conditions like `supports_flash_attn_v1() || v2 || v3` could get a bit lengthy. So, I personally prefer having a single helper function for this, though Alvaro's suggestion is also intuitive!

Happy with either approach for now :)

alvarobartt: Fair enough and makes sense yes, I just wonder if `require_fa2_capabilities` is the best flag we can use as a param? In that case maybe we can instead use an enum as `enum FlashAttentionVersion` with v1 and v2 (in the future v3), and do something like `supports_flash_attn(dtype: &DType, device: &Device, versions: Option<Vec<FlashAttentionVersion>>)`?

kozistr: It sounds good to me to use something like supports_flash_attn(...)!

michaelfeil: Wrote this code 2 months ago, for work on cpu-varlen-flash-attention, so that we don’t need all the modeling code 2x. TBH, feel free to make commits to this branch, no strong opinion.

alvarobartt:
This PR is closed in favour of https://github.com/huggingface/text-embeddings-inference/pull/825, I've included you as co-author there @michaelfeil, and thanks again for the review @kozistr! 🤗

michaelfeil:
Thank you!

PR #779, opened 2025-12-18
feat: add device id

michaelfeil:
# What does this PR do?

This adds a option to the CLI to select the device ID. 

Fixes # (issue)

## Before submitting

## Who can review?


OlivierDehaene:
Why not use the `CUDA_VISIBLE_DEVICES` env var?

michaelfeil:
@OlivierDehaene good to see you. I actually want to add py03 bindings for the backend, and wiring that though would help. Env vars can't change, so its more about adding this capability into backend.

router/src/main.rs
```diff
@@ -122,6 +122,11 @@ struct Args {
     #[clap(long, env)]
     dense_path: Option<String>,
 
+    /// The device ID to use for CUDA/Metal devices. Defaults to 0.
```
alvarobartt:
```suggestion
    /// The CUDA device ID where the model will be loaded. Defaults to 0 i.e., the first available device.
```

backends/candle/src/lib.rs
```diff
@@ -205,7 +206,7 @@ impl CandleBackend {
             #[cfg(not(feature = "cuda"))]
             Ok(Device::Cpu)
         } else if candle::utils::metal_is_available() {
-            Device::new_metal(0)
+            Device::new_metal(device_id)
```
alvarobartt: AFAIK there are no instances where you have more than one M-chip on MacOS, right?

```suggestion
            Device::new_metal(0)
```

backends/candle/src/lib.rs
```diff
@@ -188,7 +189,7 @@ impl CandleBackend {
         let device = if candle::utils::cuda_is_available() {
             #[cfg(feature = "cuda")]
             match compatible_compute_cap() {
-                Ok(true) => Device::new_cuda(0),
+                Ok(true) => Device::new_cuda(device_id),
```
alvarobartt, line 242: IMO we should try to add some sort of validation here, to ensure that the `device_id` is within the bounds of the available devices in the given instance, or is the `candle` error enough?

alvarobartt (commented):
Thank you @michaelfeil, I've left some comments!

P.S. Agree with @OlivierDehaene that using `CUDA_VISIBLE_DEVICES` would be ideal, but given that this might help with your use-case it's probably the same with other users, so I'm happy to move forward, unless Olivier thinks otherwise, I don't have a strong opinion against this change 🤗

backends/candle/src/lib.rs
michaelfeil: not at the moment, that is correct, do you want to change that?

backends/candle/src/lib.rs
michaelfeil: I feel like the candle error would be descriptive enough.

backends/candle/src/lib.rs
alvarobartt: Yes I'd revert that for the time being, unless there's a strong reason not to

alvarobartt (approved):

michaelfeil:
@alvarobartt

PR #784, opened 2025-12-22
candle upgrade to 0.9.2alpha2

michaelfeil:
- upgrades candle to latest rc
- I feel like ~5% overall improvements in performance.

https://github.com/michaelfeil/candle-extensions-fork -> Can be PR'ed and merged into huggingface's candle fork.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -28,7 +28,7 @@ jobs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v6
 
       - id: set-matrix
         env:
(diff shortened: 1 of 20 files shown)


michaelfeil:
There is some issue with the `static-linking` feature. It seems like some crates are requesting `dynamic-linking` by default. Once we are selecting `static-linking` inside this repo, but some external crates select `dynamically-linking`, we have a `cudarc` both dynamically and statically linked, which is not allowed.

alvarobartt:
Hey @michaelfeil thanks a lot for this PR, as already mentioned we're directly working with `candle` to make sure that the linking strategy doesn't default to `dynamic-linking` for neither `cudarc` nor `intel-mkl-src` so that we can simply bump `candle` without having to rely on a fork to "remove" the default linking strategy. I'll ping you on the `candle` PR later today, but once that's done IMO we can bump `candle` here too, and eventually add support for Flash Attention 3 as per your PR at #785, thanks again and sorry for the inconvenience 🤗

PR #787, opened 2025-12-22
performance-fix: notify batch task

michaelfeil:
This PR introduces a counter-based batching notification system to optimize performance by reducing excessive batching task wakeups, while fixing several code quality issues.
## Changes Made:
Core Logic Changes (core/src/infer.rs):
- Added: AtomicUsize counter-based notification system
- Modified: All inference functions to accept batch_counter: Arc<AtomicUsize> parameter
- Changed: Notification logic from notify_one() on every request to conditional notification only on last item in batch
- Updated: Function signatures for embed_pooled, embed_sparse, embed_all, predict, and classification methods
Server Layer Updates:
- gRPC Server (router/src/grpc/server.rs): Added counter initialization and passing to all inference calls
- HTTP Server (router/src/http/server.rs): Added counter initialization for both single and batch requests
## Performance Optimization:
Problem: Original implementation called notify_batching_task.notify_one() on every single request, causing excessive batching task wakeups (and processing at batch-size 1) even when processing batches of multiple requests. (aka a lot of performance lost)
Solution: Counter-based approach that only notifies when the last item in a batch is being processed:
- Single requests: Counter initialized to 1, notifies immediately
- Batch requests: Counter initialized to batch_size, notifies only on final item
## Critical Issues:
Minor Issues:
- Added proper imports for AtomicUsize and Ordering
- Consistent parameter passing across all server endpoints
Technical Details:
Counter Logic:
// Decrement counter and notify if this was the last item in the batch
if batch_counter.fetch_sub(1, Ordering::SeqCst) == 1 {
    self.notify_batching_task.notify_one();
}
- Uses fetch_sub(1) which returns the previous value
- Notifies only when counter was 1 before decrement (last item)
- Thread-safe with SeqCst ordering
Integration Points:
- Single requests: Arc::new(AtomicUsize::new(1))
- Batch requests: Arc::new(AtomicUsize::new(batch_size))
- Consistent across: gRPC, HTTP, and OpenAI-compatible endpoints

### DIFF
--- a/core/src/queue.rs
+++ b/core/src/queue.rs
@@ -46,7 +46,9 @@ impl Queue {
         max_concurrent_requests: usize,
     ) -> Self {
         // Create channels
-        let (queue_sender, queue_receiver) = mpsc::channel(max_concurrent_requests);
+        // The queue rarely fails to send the QueueCommand due to a lack of buffer size.
+        // So, naively increasing the buffer size to twice than `max_concurrent_requests` to prevent the failure temporarily
+        let (queue_sender, queue_receiver) = mpsc::channel(2 * max_concurrent_requests);
 
         // Launch background queue task
         std::thread::spawn(move || {
--- a/router/src/http/server.rs
+++ b/router/src/http/server.rs
@@ -123,7 +123,7 @@ async fn predict(
                               permit: Option<OwnedSemaphorePermit>,
                               _batch_counter: Option<Arc<AtomicUsize>>| async move {
         let permit = match permit {
-            None => infer.acquire_permit().await,
+            None => infer.try_acquire_permit().map_err(ErrorResponse::from)?,
             Some(permit) => permit,
         };
 
@@ -361,7 +361,7 @@ async fn rerank(
                              truncate: bool,
                              infer: Infer,
                              batch_counter: Option<Arc<AtomicUsize>>| async move {
-        let permit = infer.acquire_permit().await;
+        let permit = infer.try_acquire_permit().map_err(ErrorResponse::from)?;
 
         let response = infer
             .predict(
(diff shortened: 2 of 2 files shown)


router/src/grpc/server.rs
```diff
@@ -264,6 +268,7 @@ impl TextEmbeddingsService {
         truncation_direction: tokenizers::TruncationDirection,
         raw_scores: bool,
         permit: OwnedSemaphorePermit,
+        batch_counter: Arc<AtomicUsize>,
```
kozistr: Looks like `batch_counter` here might be an optional type since `None` is passed to it at line 743!

```suggestion
        batch_counter: Optional<Arc<AtomicUsize>>,
```

Also, since `batch_counter` doesn't seem to be used here, maybe it'd make sense to rename it like this?

```suggestion
        _batch_counter: Optional<Arc<AtomicUsize>>,
```

michaelfeil: thanks, fixed

michaelfeil:
Complies on both, grpc and http server now.

```
 cargo build --release --bin text-embeddings-router -F candle-cuda -F dynamic-linking -F grpc --no-default-features
   Compiling text-embeddings-router v1.8.3 (/workspace/model-performance/michaelfeil1209/text-embeddings-inference/router)
    Finished `release` profile [optimized] target(s) in 2m 40s
root@michaelfeil1209-dev-pod-h100-0 /workspace/model-performance/michaelfeil1209/text-embeddings-inference  $ cargo build --release --bin text-embeddings-router -F candle-cuda -F dynamic-linking -F http --no-default-features
    Blocking waiting for file lock on build directory
   Compiling text-embeddings-router v1.8.3 (/workspace/model-performance/michaelfeil1209/text-embeddings-inference/router)
    Finished `release` profile [optimized] target(s) in 2m 53s
root@michaelfeil1209-dev-pod-h100-0 /workspace/model-performance/michaelfeil1209/text-embeddings-inference  $ 
```

PR #789, opened 2025-12-26
feat/jina embeddings v4

kang-raison:
# What does this PR do?

This PR adds full support for **Jina Embeddings v4** across routing, backend handling, and documentation.

### Summary of changes

- **Model detection & routing**
  - Detects `JinaEmbeddingsV4Model` and `text_config` fields in `config.json`
  - Applies `sliding_window` logic to respect max input length
  - Treats Jina v4 as an **embedding model with mean pooling by default** in the router

- **Python backend support**
  - Adds a Python backend handler for **Jina v4** (remote code + LoRA support)
  - Returns `single_vec_emb`
  - Supports automatic task selection, with optional override via `JINA_V4_TASK`

- **Build & test improvements**
  - Improves default dtype handling for **gRPC-only builds**
  - Fixes tokenizer test to use the **Tokio API**

- **Documentation**
  - Documents Jina v4 usage in the README
  - Adds Jina v4 to the supported models list

Fixes # (no issue)

## Before submitting

- [ ] Was this discussed/approved via a GitHub issue or the [forum](https://discuss.huggingface.co/)? (No)
- [x] Did you make sure to update the documentation with your changes? (README + supported_models)
      - Workspace tests  
      - Jina integration tests  
      - Tokenizer test updated to use Tokio API

## Who can review?

@Narsil  
@alvarobartt

### DIFF
--- a/backends/candle/src/lib.rs
+++ b/backends/candle/src/lib.rs
@@ -31,8 +31,10 @@ use crate::models::{
 use crate::models::{
     FlashBertModel, FlashDistilBertModel, FlashGTEModel, FlashJinaBertModel,
     FlashJinaCodeBertModel, FlashMistralModel, FlashModernBertModel, FlashNomicBertModel,
-    FlashQwen2Model, FlashQwen3Model,
+    FlashQwen2Model, FlashQwen3Model, LoraWeights,
 };
+#[cfg(feature = "cuda")]
+use std::{env, fs};
 
 /// This enum is needed to be able to differentiate between jina models that also use
 /// the `bert` model type and valid Bert models.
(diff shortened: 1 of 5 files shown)


backends/src/dtype.rs
```diff
@@ -12,7 +12,6 @@ pub enum DType {
         all(feature = "candle", not(feature = "accelerate"))
     ))]
     Float16,
-    #[cfg(any(feature = "python", feature = "candle", feature = "ort"))]
```
michaelfeil, line 15: should not be removed.

backends/python/server/text_embeddings_server/models/jina_v4.py
```diff
@@ -0,0 +1,81 @@
...
+from opentelemetry import trace
+from transformers import AutoModel
+
+from text_embeddings_server.models import Model
+from text_embeddings_server.models.types import Embedding, PaddedBatch
+
+tracer = trace.get_tracer(__name__)
+
+
+class JinaV4Model(Model):
```
michaelfeil: why not implement JinaV4Model in candle? Its much faster than torch here, actually (because you have to use PaddedBatch)

PR #799, opened 2026-01-08
Feature: Adding Configurable Batch Channel Capacity for Pipeline Parallelism

### LINKED ISSUE #798: Configurable Batch Channel Capacity for Pipeline Parallelism
### Feature request

Currently, the batch processing channel in `core/src/infer.rs` has a hardcoded capacity, limiting the ability to achieve pipeline parallelism between batch formation and inference. This creates a bottleneck where:

1. **Sequential Processing**: Only one batch can be in-flight at a time, forcing the batching task to wait for inference to complete before forming the next batch
2. **Reduced Throughput**: Under high concurrency, requests queue up waiting for batches to be processed sequentially
3. **No Tuning**: Users cannot optimize the latency/throughput trade-off for their specific workload

### Motivation

With capacity=1, the batching task must wait for the backend to finish processing before it can send the next batch, preventing pipeline parallelism.

Key Metrics to Measure

- **Queue Time**: Time requests spend waiting in the batching queue
- **Batch Size**: Number of requests processed together
- **Throughput**: Requests per second
- **Latency**: End-to-end request time

### Your contribution

Will submit PR with proposed change and benchmarking numbers.

jovsa:
# What does this PR do?

Currently, the batch processing channel in `core/src/infer.rs` has a hardcoded capacity, limiting the ability to achieve pipeline parallelism between batch formation and inference.

This PR, adds a configurable `--batch-channel-capacity` parameter that allows users to tune how many batches can be queued for processing simultaneously.

Fixes: https://github.com/huggingface/text-embeddings-inference/issues/798

## Before submitting

- [X] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [X] Did you read the [contributor guideline](https://github.com/huggingface/text-embeddings-inference/blob/main/CONTRIBUTING.md)?

## Who can review?

### DIFF
--- a/core/src/infer.rs
+++ b/core/src/infer.rs
@@ -27,11 +27,12 @@ impl Infer {
         queue: Queue,
         max_concurrent_requests: usize,
         backend: Backend,
+        batch_channel_capacity: usize,
     ) -> Self {
         let notify_batching_task = Arc::new(Notify::new());
 
-        // Bound channel to 1 to be able to prefetch one batch
-        let (embed_sender, embed_receiver) = mpsc::channel(1);
+        // Channel capacity controls how many batches can be queued for processing.
+        let (embed_sender, embed_receiver) = mpsc::channel(batch_channel_capacity);
 
         // Batching task
         tokio::spawn(batching_task(
--- a/router/src/lib.rs
+++ b/router/src/lib.rs
@@ -66,6 +66,7 @@ pub async fn run(
     otlp_service_name: String,
     prometheus_port: u16,
     cors_allow_origin: Option<Vec<String>>,
+    batch_channel_capacity: usize,
 ) -> Result<()> {
     let model_id_path = Path::new(&model_id);
     let (model_root, api_repo) = if model_id_path.exists() && model_id_path.is_dir() {
@@ -316,7 +317,7 @@ pub async fn run(
     );
 
     // Create infer task
-    let infer = Infer::new(tokenization, queue, max_concurrent_requests, backend);
+    let infer = Infer::new(tokenization, queue, max_concurrent_requests, backend, batch_channel_capacity);
 
     // Endpoint info
     let info = Info {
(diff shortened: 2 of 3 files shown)


michaelfeil:
@jovsa the PR you actually want to do is like 500 LOC, and this change is just not gonna cut it. 

You need to: 
- move all mover threads that are spawned and are syncrounous
- for that you need to clone() but Candle backend is running in pinned memory, you can't clone. 
- multiple backends running at the same time will interfer each other (two concurrent matmuls run at less! than 50% of the speed) The reason for that is L2 eviction. 

Backends need to be cloneable, as result this PR does not do anything to performance.

However it addtionally is hindering HTTP drop based cancellation. By pre-batching more than 1 request and pushing it into a queue, there is no option to cancel . This is a regression / bug.

michaelfeil:
@jovsa I beleive your performance numbers are fully made up. Have you used AI to write this PR? Have you validated the perf at all?

michaelfeil:
Let me help answer this:

With 1-slot channel: The batching task only creates a new batch when the GPU is ready for the next one. This ensures optimal resource utilization.
With larger capacity: We might create multiple batches while the GPU is still processing the first one, leading to:
- Memory waste: Multiple large batches allocated simultaneously
- Stale data: Batches might contain requests that have timed out
- Poor responsiveness: System can't quickly adapt to changing load patterns

Scenario 1: Sudden traffic spike
- Current: Backend processes current batch, batching task blocks, new requests queue efficiently
- Larger channel: Batching task creates 2+ batches, GPU takes longer to adapt because you run things at low batch-size.
Scenario 2: GPU memory pressure
- Current: Backend slows down, batching task naturally throttles via channel backpressure
- Larger channel: Batching task continues creating batches, exacerbating memory issues
The "Prefetch" Design Intent 📝
The comment says "Bound channel to 1 to be able to prefetch one batch" - this is actually about allowing exactly one batch to be prepared while the backend is processing the previous one. It's not about having multiple batches ready simultaneously.

Assumption: Popping batches takes >>0.5ms (I timed it). Even a single token on B200 on bert small is on the order of 1-2ms due to kernel launch overhead.

I hope this answers your question.

PR #801, opened 2026-01-10
Fix `PredictInput::Batch` and expect pairs as nested tuples

### LINKED ISSUE #747: /predict does not work for batches for sequence classification
### System Info

It seems that whenever a set of inputs as batch is being passed in to `/predict` for sequence classification, only the first result is returned.

### Information

- [x] Docker
- [x] The CLI directly

### Tasks

- [ ] My own modifications

### Reproduction

```
text-embeddings-router --model-id HuggingFaceFW/fineweb-edu-classifier --port 8080
```

```
~/dev/open-source-help/text-embeddings-inference main = *4 ?8 > curl -XPOST localhost:8080/predict -H 'Content-Type: application/json' -d '{"inputs": "Butterflies cute"}' | jq                 py 3.11  12:29:07
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    70  100    40  100    30    696    522 --:--:-- --:--:-- --:--:--  1228
[
  {
    "score": 0.50506574,
    "label": "LABEL_0"
  }
]

~/dev/open-source-help/text-embeddings-inference main = *4 ?8 > curl -XPOST localhost:8080/predict -H 'Content-Type: application/json' -d '{"inputs": ["Butterflies cute", "Where can I find the one piece?"]}' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   106  100    39  100    67   1324   2274 --:--:-- --:--:-- --:--:--  3655
[
  {
    "score": 0.5077051,
    "label": "LABEL_0"
  }
]

~/dev/open-source-help/text-embeddings-inference main = *4 ?8 > curl -XPOST localhost:8080/predict -H 'Content-Type: application/json' -d '{"inputs": ["Butterflies cute", "Where can I find the one piece?", "How about now"]}'     
Failed to deserialize the JSON body into the target type: inputs: invalid length 3, expected a string, a pair of strings [string, string] or a batch of mixed strings and pairs [[string], [string, string], ...] at line 1 column 83%   
```

### Expected behavior

TEI should either throw an error `/predict` endpoint when size is 2 and model does not do pair classification.

alvarobartt:
# What does this PR do?

This PR allows a `Vec<String>` with length > 2 to be provided as `PredictInput::Batch`, given that it was preventing some routes as `/predict` on processing with more than 2 sequences as those were considered "pairs". 

This PR represents a breaking change on how the `/predict` method works, as before this PR pairs could be provided as `[a,b]` or `[[a,b]]`, but the fact that `[a,b]` was allowed was preventing the inputs to be a batch of sequences, given that it was limited to two sequences, and then interpreted as a pair rather than as two individual sequences.

So on, from now on, only the following are allowed:
- Single sequences as `"inputs": "a"`
- Batch of sequences as `"inputs":["a", ...]` (allows from 1 to N, defined by the `--max-client-batch-size` with a default value of 32)
- Batch of single sequences and/or pairs as `"inputs":[["a"], ...]` and `"inputs":[["a","b"], ...]`, respectively; or combined as `"inputs":[["a"], ["b", "c"], ...]`.

> [!WARNING]
> It's a breaking change being introduced without a deprecation warning or prior notice, given that treating a batch of sequences as a pair, which implies that the strings there are concatenated into a single sequence, and limiting the length of batches of sequences to 2; can be misleading and/or incorrect.
>
> Since before calling `/predict` with `{"inputs":["I love tennis", "I hate tennis"]}` was not running the forward pass on a batch of `["I love tennis", "I hate tennis"]`, but rather running on a single sequence formed by the concatenation of both sequences in the batch, leading then to a single classification score, rather than an independent score per sequence; and also preventing the user from running with larger batches.

Fixes #747 

## Before submitting

- [X] Did you read the [contributor guideline](https://github.com/huggingface/text-embeddings-inference/blob/main/CONTRIBUTING.md)?
- [X] Was this discussed/approved via a GitHub issue or the [forum](https://discuss.huggingface.co/)? Please add a link to it if that's the case.

## Who can review?

### DIFF
--- a/router/src/http/types.rs
+++ b/router/src/http/types.rs
@@ -3,9 +3,20 @@ use serde::de::{SeqAccess, Visitor};
 use serde::{de, Deserialize, Deserializer, Serialize};
 use serde_json::json;
 use std::fmt::Formatter;
+use std::sync::OnceLock;
 use text_embeddings_core::tokenization::EncodingInput;
 use utoipa::openapi::{RefOr, Schema};
 use utoipa::ToSchema;
+
+static MAX_CLIENT_BATCH_SIZE: OnceLock<usize> = OnceLock::new();
+
+pub(crate) fn set_max_client_batch_size(size: usize) {
+    MAX_CLIENT_BATCH_SIZE.get_or_init(|| size);
+}
+
+fn get_max_client_batch_size() -> usize {
+    *MAX_CLIENT_BATCH_SIZE.get().unwrap_or(&32)
+}
 
 #[derive(Debug)]
 pub(crate) enum Sequence {
@@ -92,7 +103,7 @@ impl<'de> Deserialize<'de> for PredictInput {
                     // Flat array of strings: ["a", "b", "c", ...]
                     // Always treated as a batch of single sequences
                     Internal::Single(first_string) => {
-                        let mut batch = Vec::with_capacity(32);
+                        let mut batch = Vec::with_capacity(get_max_client_batch_size());
                         batch.push(Sequence::Single(first_string));
 
                         while let Some(s) = seq.next_element::<String>()? {
@@ -104,7 +115,7 @@ impl<'de> Deserialize<'de> for PredictInput {
                     // Nested array: [["a"], ["a", "b"], ...]
                     // Batch of singles and/or pairs
                     Internal::Multiple(first_vec) => {
-                        let mut batch = Vec::with_capacity(32);
+                        let mut batch = Vec::with_capacity(get_max_client_batch_size());
                         batch.push(sequence_from_vec(first_vec)?);
 
                         while let Some(vec) = seq.next_element::<Vec<String>>()? {
--- a/router/src/http/types.rs
+++ b/router/src/http/types.rs
@@ -57,9 +57,8 @@ impl<'de> Deserialize<'de> for PredictInput {
             fn expecting(&self, formatter: &mut Formatter) -> std::fmt::Result {
                 formatter.write_str(
                     "a string, \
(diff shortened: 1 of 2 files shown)


router/src/http/types.rs
```diff
@@ -73,63 +73,47 @@ impl<'de> Deserialize<'de> for PredictInput {
...
-                        if seq.next_element::<String>()?.is_some() {
-                            // Error as we do not accept > 2 elements
-                            return Err(de::Error::invalid_length(3, &self));
+                    .ok_or_else(|| de::Error::invalid_length(0, &self))?;
+
+                match first {
+                    // Flat array of strings: ["a", "b", "c", ...]
+                    // Always treated as a batch of single sequences
+                    Internal::Single(first_string) => {
+                        let mut batch = Vec::with_capacity(32);
```
michaelfeil: why 32?

alvarobartt: I also thought about it, I guess that's because the default value for `--max-client-batch-size` is 32, so it's "hardcoded" there (not something that I introduced in this PR). But yes, we should find a way to remove the hardcoded value and rather use whatever value is provided as `--max-client-batch-size`, I'll look into it before merging the PR.

alvarobartt: https://github.com/huggingface/text-embeddings-inference/pull/801/commits/024db5a5aa9f0e135cdcadb0d0e80a6bec76aac7 🤗

PR #860, opened 2026-04-09
feat: ROCm flash-attn varlen, triton layer norm, and AMD Dockerfile

Abdennacer-Badaoui:
Adds flash attention varlen, triton-based layer norm/rms norm via `kernels-community/triton-layer-norm`, and a dedicated `Dockerfile-amd` with CI matrix entry for MI300/MI325 GPUs.


backends/python/server/text_embeddings_server/models/flash_bert.py
```diff
@@ -9,11 +9,19 @@
...
+from text_embeddings_server.utils.flash_attn import attention, ROCM_HAS_FA_VARLEN
 from text_embeddings_server.utils.device import is_rocm, use_ipex
 
 tracer = trace.get_tracer(__name__)
 
+_triton_layer_norm = None
+if is_rocm():
+    try:
+        from kernels import get_kernel as _get_kernel
+        _triton_layer_norm = _get_kernel("kernels-community/triton-layer-norm")
```
alvarobartt: Can we specify the revision / version here?

backends/python/server/text_embeddings_server/models/flash_mistral.py
```diff
@@ -10,11 +10,19 @@
...
+from text_embeddings_server.utils.flash_attn import attention, ROCM_HAS_FA_VARLEN
 from text_embeddings_server.utils.device import is_rocm
 
 tracer = trace.get_tracer(__name__)
 
+_triton_layer_norm = None
+if is_rocm():
+    try:
+        from kernels import get_kernel as _get_kernel
+        _triton_layer_norm = _get_kernel("kernels-community/triton-layer-norm")
```
alvarobartt: Same here with the revision / version

docs/source/en/amd_gpu.md
```diff
@@ -16,19 +16,45 @@ rendered properly in your Markdown viewer.
...
+
+docker run \
+  --device /dev/kfd --device /dev/dri \
+  --group-add video \
+  --ipc=host \
+  -p 8080:80 \
+  -v $volume:/data \
+  --pull always \
+  ghcr.io/huggingface/text-embeddings-inference:rocm-latest \
+  --model-id $model --dtype bfloat16
```
alvarobartt, line 41: Since it just supports at most a single accelerator, could you instead specify the `--device /dev/dri/renderD128` instead of `--device /dev/dri` or rather using the environment variable `ROCR_VISIBLE_DEVICES` (as per https://rocm.docs.amd.com/en/latest/conceptual/gpu-isolation.html#rocr-visible-devices)?

docs/source/en/amd_gpu.md
```diff
@@ -16,19 +16,45 @@ rendered properly in your Markdown viewer.
...
+```
+
+Then test it:
+
+```shell
+curl http://localhost:8080/embed \
+    -X POST \
+    -H 'Content-Type: application/json' \
+    -d '{"inputs": "What is Deep Learning?"}'
```
alvarobartt: Might be better to use the OpenAI-compatible API at `/v1/embeddings` instead

Dockerfile-amd
```diff
@@ -0,0 +1,103 @@
+# ── Rust builder ─────────────────────────────────────────────────────────────
```
alvarobartt:
```suggestion
```

Dockerfile-amd
```diff
@@ -0,0 +1,103 @@
...
+WORKDIR /usr/src
+COPY backends backends
+COPY proto proto
+COPY backends/python/server/pyproject.toml backends/python/server/pyproject.toml
+COPY backends/python/server/requirements-amd.txt backends/python/server/requirements.txt
+
+RUN cd backends/python/server && \
+    make install
+
+# ── Final images ──────────────────────────────────────────────────────────────
```
alvarobartt:
```suggestion
```

Dockerfile-amd
```diff
@@ -0,0 +1,103 @@
...
+
+FROM builder AS grpc-builder
+
+COPY proto proto
+
+RUN --mount=type=secret,id=actions_results_url,env=ACTIONS_RESULTS_URL \
+    --mount=type=secret,id=actions_runtime_token,env=ACTIONS_RUNTIME_TOKEN \
+    cargo build --release --bin text-embeddings-router -F grpc -F python --no-default-features && sccache -s
+
+# ── ROCm Python runtime ───────────────────────────────────────────────────────
```
alvarobartt:
```suggestion
```

.github/workflows/matrix.json
```diff
@@ -124,5 +124,14 @@
     "extraBuildArgs": "PLATFORM=hpu",
     "grpc": true,
     "dockerfile": "Dockerfile-intel"
+  },
+  {
+    "name": "rocm-mi300-mi325",
+    "imageNamePrefix": "rocm-",
+    "runOn": "always",
+    "sccache": true,
+    "extraBuildArgs": "GPU_ARCHS=gfx942",
```
alvarobartt: Should we also build with `GPU_ARCHS=gfx950`, or does this image works for those too?

.github/workflows/matrix.json
```diff
@@ -124,5 +124,14 @@
     "extraBuildArgs": "PLATFORM=hpu",
     "grpc": true,
     "dockerfile": "Dockerfile-intel"
+  },
+  {
+    "name": "rocm-mi300-mi325",
```
alvarobartt: Maybe this fits better here?


```suggestion
    "name": "rocm-gfx942",
```

alvarobartt (dismissed):
Hey @Abdennacer-Badaoui, thanks a lot, left some comments / thoughts, feel free to apply or discard where applicable 🤗

docs/source/en/amd_gpu.md
Abdennacer-Badaoui: Yes I agree. Let me change this

.github/workflows/matrix.json
Abdennacer-Badaoui: yes, gfx950 already works, no separate build needed

.github/workflows/matrix.json
Abdennacer-Badaoui: I went with rocm-gfx942-gfx950 rather than just rocm-gfx942, since the prebuilt flash-attn wheel (whl/gfx942-gfx950/...) covers both

backends/python/server/text_embeddings_server/utils/flash_attn.py
```diff
@@ -16,8 +16,21 @@
...
+        ROCM_HAS_FA_VARLEN = True
+        HAS_FLASH_ATTN_V2 = True
+        logger.info("ROCm: flash_attn found, using flash_attn_varlen_func")
+    except ImportError:
+        ROCM_HAS_FA_VARLEN = False
+        HAS_FLASH_ATTN_V2 = True  # still True — rocm_attn SDPA fallback handles it
+        logger.warning(
+            "ROCm: flash_attn not found, falling back to SDPA. "
+            "Install flash-attn for better performance."
+        )
```
alvarobartt, line 33:
```suggestion
    HAS_FLASH_ATTN_V2 = True
    
    try:
        from flash_attn import flash_attn_varlen_func as _fa_varlen_func
        ROCM_HAS_FA_VARLEN = True
        logger.info("ROCm: flash_attn found, using flash_attn_varlen_func")
    except ImportError:
        ROCM_HAS_FA_VARLEN = False
        logger.warning(
            "ROCm: flash_attn not found, falling back to SDPA. "
            "Install flash-attn for better performance."
        )
```

alvarobartt: In any case, given that it's true on both branches here, does it makes sense to replace the `if HAS_FLASH_ATTN_V2:` below with something like `if use_ipex and HAS_FLASH_ATTN_V2:` and then for ROCm do something like `if use_rocm:` and internally handle the branches for `ROCM_HAS_FA_VARLEN`?

backends/python/server/text_embeddings_server/utils/flash_attn.py
alvarobartt: In my earlier comment, I mean keeping this if only for IPEX, and a separate one for ROCm given that `HAS_FLASH_ATTN_V2` is always true for ROCm.

Dockerfile-amd
```diff
@@ -0,0 +1,100 @@
...
+RUN python3 -m pip install --no-cache-dir "${FLASH_ATTN_WHEEL}"
+
+WORKDIR /usr/src
+COPY backends backends
+COPY proto proto
+COPY backends/python/server/pyproject.toml backends/python/server/pyproject.toml
+COPY backends/python/server/requirements-amd.txt backends/python/server/requirements.txt
+
+RUN cd backends/python/server && \
+    make install
```
alvarobartt: Opinionated nit, feel free to discard if you don't like it

```suggestion
RUN cd backends/python/server && make install
```

Dockerfile-amd
```diff
@@ -0,0 +1,100 @@
...
+    build-essential \
+    git \
+    python3-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install flash-attention from AMD's prebuilt ROCm wheels (no source build).
+# The wheel is built for gfx942 (MI300X / MI300A) and gfx950 (MI350) and is
+# pinned to match the ROCm / Python / PyTorch ABI of the base image.
+ARG FLASH_ATTN_WHEEL="https://rocm.frameworks-devreleases.amd.com/whl/gfx942-gfx950/flash_attn-2.8.3-cp310-cp310-linux_x86_64.whl"
+RUN python3 -m pip install --no-cache-dir "${FLASH_ATTN_WHEEL}"
```
alvarobartt: Hmm do we really need this as an `ARG`?

```suggestion
RUN python3 -m pip install --no-cache-dir "https://rocm.frameworks-devreleases.amd.com/whl/gfx942-gfx950/flash_attn-2.8.3-cp310-cp310-linux_x86_64.whl"
```

Dockerfile-rocm
alvarobartt, line 1: Maybe another opinionated nit, but do you mind renaming this to `Dockerfile-rocm` instead?

.github/workflows/matrix.json
```diff
@@ -124,5 +124,13 @@
...
     "grpc": true,
     "dockerfile": "Dockerfile-intel"
+  },
+  {
+    "name": "rocm-gfx942-gfx950",
+    "imageNamePrefix": "rocm-",
+    "runOn": "always",
+    "sccache": true,
+    "grpc": true,
+    "dockerfile": "Dockerfile-amd"
```
alvarobartt:
```suggestion
    "dockerfile": "Dockerfile-rocm"
```

Assuming that the `Dockerfile-amd` is renamed to `Dockerfile-rocm`

alvarobartt (dismissed):

backends/python/server/text_embeddings_server/utils/flash_attn.py
Abdennacer-Badaoui: Yes, i think this is much more cleaner.

Dockerfile-amd
Abdennacer-Badaoui: Not really, it was just for cleanliness :)

backends/python/server/text_embeddings_server/models/jinaBert_model.py
```diff
@@ -12,9 +12,20 @@
...
 
 tracer = trace.get_tracer(__name__)
 
+_triton_layer_norm = None
+if is_rocm():
+    try:
+        from kernels import get_kernel as _get_kernel
+        _triton_layer_norm = _get_kernel(
+            "kernels-community/triton-layer-norm", revision="v0.1.0"
+        )
```
alvarobartt: Hey @Abdennacer-Badaoui after carefully reviewing it, I'm concerned about downloading kernels on runtime, do you think we could download the kernel within the Dockerfile so that the container doesn't pull kernels on runtime? I know that mounting the cache with the kernel would work i.e., prevent the egress from within the container to the Hub, but still I'd like if the container can be network isolated and not try to call the Hub, thoughts? cc @pagezyhf for visibility

Abdennacer-Badaoui: Good call @alvarobartt. The kernel is now baked into the image at build time (kernels download into `KERNELS_CACHE=/kernels`) and pinned via a committed `kernels.lock`. At runtime the code uses `load_kernel(..., lockfile=...)` with `local_files_only=True`, so the container never calls the Hub and stays network-isolatable.

backends/python/server/text_embeddings_server/utils/kernels.py
```diff
@@ -0,0 +1,31 @@
...
+    try:
+        # `load_kernel` loads a pre-downloaded, locked kernel from the local
+        # cache with `local_files_only=True`, so the container never reaches
+        # out to the Hub at runtime. The kernel is fetched at build time via
+        # `kernels download` (see Dockerfile-rocm) into `KERNELS_CACHE`.
+        from kernels import load_kernel
+
+        _triton_layer_norm = load_kernel(
+            "kernels-community/triton-layer-norm", lockfile=_LOCKFILE
+        )
```
alvarobartt: Will this pull from the Hub when we don't run in a container? i.e., if I clone `text-embeddings-inference` and run with `cargo r --release ...` for AMD, would that still pull the kernels? Or does this mean that this is assuming that the kernel exists already?

Abdennacer-Badaoui: As discussed offline, i added get_kernel fallback

PR #869, opened 2026-05-26
refactor: streamline tensor initialization and rearrange struct fields for clarity

Unmesh100:
1. Bug 1 (softmax fix): The change is trivial - removing .abs() on both sides. This is a standard softmax numerical stability fix. Compilation passes with cargo check -p text-embeddings-router which includes the core crate. No behavioral tests needed for this - it's a well-understood mathematical correction.
2. Bug 2 (Qwen3 CLS pooling): The change removes an incorrect conditional branch that was selecting diagonal elements. Now it always selects column 0 (CLS token). Compilation passes. The existing snapshot tests in backends/candle/tests/test_qwen3.rs would catch any major regression in Qwen3 embeddings, but they download models from HuggingFace which is not feasible here.
3. Bug 3 (shutdown deadlock): The field reordering is a zero-behavior-change fix (same fields, just different drop order). cargo check and cargo test --no-run both compile successfully with the new field order.

### DIFF
--- a/backends/candle/src/models/qwen3.rs
+++ b/backends/candle/src/models/qwen3.rs
@@ -646,15 +646,11 @@ impl Qwen3Model {
                             &self.device,
                         )?;
 
-                        let cls_indices = if has_raw_requests {
-                            Tensor::zeros(
-                                batch.pooled_indices.len(),
-                                candle::DType::U32,
-                                &self.device,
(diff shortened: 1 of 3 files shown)


Unmesh100:
@Narsil can you please review it, thanks.

PR #884, opened 2026-06-25
feat: allow controlling startup warmup tokens

malaiwah:
## What does this PR do?

Adds a `--warmup-tokens` / `WARMUP_TOKENS` option to control the explicit startup warmup pass.

- Default behavior is unchanged:
  - padded backends use `min(max_input_length, max_batch_tokens)`
  - non-padded backends use `max_batch_tokens`
- `--warmup-tokens 0` skips the explicit warmup pass.
- `--warmup-tokens N` runs a smaller synthetic warmup with `N` tokens.
- Positive values above `--max-batch-tokens` are rejected to avoid startup-only shapes larger than production batching limits.
- HPU and ROCm keep their existing backend-specific bucket warmup for positive values; `0` still skips before that warmup is entered.

This addresses #819, where CPU/ORT Qwen3 ONNX startup was dominated by the warmup pass even after lowering `--max-batch-tokens`.

## Why

Warmup is valuable on GPU because it exercises production batch sizes and can trigger kernel compilation / memory allocation. On CPU spot deployments, especially ORT/ONNX, the same synthetic forward pass can dominate cold start time.

This lets users choose a smaller smoke-test warmup or skip the explicit warmup pass entirely without also lowering `--max-batch-tokens` and changing serving limits.

## Validation

Static/local checks:

```bash
cargo test -p text-embeddings-backend --no-default-features --features ort,clap
cargo test -p text-embeddings-backend --no-default-features --features candle,clap
cargo check -p text-embeddings-router --no-default-features --features ort,http
cargo check -p text-embeddings-router --no-default-features --features candle,http
cargo run -p text-embeddings-router --no-default-features --features ort,http -- --help
git diff --check
```

Runtime timing on macOS/arm64 CPU, release-built ORT/http router, using the model from #819:

```text
Svenni551/Qwen3-Embedding-0.6B-ONNX-INT8
```

Note: current TEI could not start directly from that repo config because the config contains GPT-style alias fields (`n_embd`, `n_layer`, `n_head`) that parse as duplicate aliases for existing Qwen3 fields. I used a temporary local copy with only those alias keys removed; the ONNX weights and tokenizer were unchanged.

Measured process start until `GET /health` succeeded:

| Setup | Default warmup | `--warmup-tokens 0` | `--warmup-tokens 64` |
|---|---:|---:|---:|
| `--max-batch-tokens 1024` | `3.163s` | `1.049s` | `1.590s` |
| default `16384` | still warming after ~4 min, stopped manually | `1.594s` | `1.590s` |

## Before submitting

- [x] Was this discussed/approved via a GitHub issue or the [forum](https://discuss.huggingface.co/)? See #819.
- [x] Did you make sure to update the documentation with your changes?
- [x] Did you write any new necessary tests? If applicable, did you include or update the `insta` snapshots? Added unit tests for warmup token selection and validation; no snapshots needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,6 +321,13 @@ Options:
           [env: PROMETHEUS_PORT=]
           [default: 9000]
 
+      --warmup-tokens <WARMUP_TOKENS>
+          Control the number of tokens used for model warmup (a synthetic inference pass that can pre-allocate memory and trigger kernel compilation).
+
+          Defaults to `min(max_input_length, max_batch_tokens)` on CPU and `max_batch_tokens` on GPU. Set to `0` to skip the explicit warmup pass for fast cold-start scenarios like CPU spot instances. Positive values are ignored by HPU and ROCm backends, which use backend-specific warmup.
+
+          [env: WARMUP_TOKENS=]
+
       --cors-allow-origin <CORS_ALLOW_ORIGIN>
           Unused for gRPC servers
 
--- a/backends/src/lib.rs
+++ b/backends/src/lib.rs
@@ -75,6 +75,23 @@ fn is_rocm() -> bool {
     }
 }
 
+fn effective_warmup_tokens(
+    max_input_length: usize,
+    max_batch_tokens: usize,
+    padded_model: bool,
+    warmup_tokens: Option<usize>,
+) -> Result<Option<usize>, BackendError> {
+    match warmup_tokens {
+        Some(0) => Ok(None),
+        Some(tokens) if tokens > max_batch_tokens => Err(BackendError::Start(format!(
+            "warmup_tokens ({tokens}) exceeds max_batch_tokens ({max_batch_tokens}); set `--warmup-tokens` to at most `--max-batch-tokens` or increase `--max-batch-tokens`"
+        ))),
+        Some(tokens) => Ok(Some(tokens)),
+        None if padded_model => Ok(Some(max_input_length.min(max_batch_tokens))),
+        None => Ok(Some(max_batch_tokens)),
+    }
+}
+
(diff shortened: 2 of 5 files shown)


malaiwah:
CUDA validation added on an RTX 5090 host.

Environment:

```text
Host: Ubuntu 24.04, NVIDIA GeForce RTX 5090
Driver: 595.58.03; nvidia-smi reports CUDA 13.2
Container: podman + docker.io/nvidia/cuda:12.8.0-devel-ubuntu24.04
CUDA_COMPUTE_CAP=120
```

I used a validation-only `RUSTFLAGS='--cfg feature="dynamic-linking"'` in the container so `cudarc` links dynamically against the CUDA runtime available there. No repository files were changed for this.

Backend warmup tests with CUDA enabled:

```bash
cargo test -p text-embeddings-backend \
  --no-default-features \
  --features candle,cuda,clap
```

Result:

```text
test result: ok. 5 passed; 0 failed
```

Router CUDA build check, using the backend's transitive CUDA feature because `text-embeddings-router` does not expose a direct `cuda` feature:

```bash
cargo check -p text-embeddings-router \
  --no-default-features \
  --features candle,http,text-embeddings-backend/cuda
```

Result:

```text
Finished `dev` profile [unoptimized + debuginfo] target(s) in 3m 11s
```

CLI surface check:

```bash
cargo run -p text-embeddings-router \
  --no-default-features \
  --features candle,http,text-embeddings-backend/cuda \
  -- --help
```

Result: help output includes the new option:

```text
--warmup-tokens <WARMUP_TOKENS>
```

So the warmup-token tests and router CLI compile cleanly with CUDA enabled.

Issue #141, opened 2024-01-30
Support for BAAI/bge-m3 model


xusenlinzy:
### Feature request

[BGE-M3](https://huggingface.co/BAAI/bge-m3), which is distinguished for its versatility in Multi-Functionality, Multi-Linguality, and Multi-Granularity.

+ Multi-Functionality: It can simultaneously perform the three common retrieval functionalities of embedding model: dense retrieval, multi-vector retrieval, and sparse retrieval.

+ Multi-Linguality: It can support more than 100 working languages.

+ Multi-Granularity: It is able to process inputs of different granularities, spanning from short sentences to long documents of up to 8192 tokens.

### Motivation

Support for BAAI/bge-m3 model

### Your contribution

Support for BAAI/bge-m3 model

jonmach:
This seems to work already

spttt:
> This seems to work already

really?  Is it supported now?

jonmach:
I didn't realise it wasn't supported.  I just loaded it up and it worked.

LLukas22:
Generating normal `dense` embeddings works fine because `bge-m3` is just a regular `XLM-Roberta` model. 

The problem is there's no way to use the `sparse` or `colbert` features of this model because they need different linear heads on the model's unpooled output, and right now, it seems like there's no way to get TEI to give back the `last_hidden_state` of the model, which you need to use those heads.

Maybe TEI could be expanded to have an endpoint/option, that returns the raw output of an embeding model and maybe the masks, as those are needed if we manually want to process the outputs on the client side. That way the bulk of the processing can happen on the TEI side and only the last linear layers need to be applied on the client side.

edwardzjl:
@jonmach Hi, could you please share how do you start TEI with bge-m3? I'm having trouble starting the service.

TEI seems not downloding the pooling config file (`1_Pooling/config.json`), which apears in the model repo.

```console
$ text-embeddings-router --model-id BAAI/bge-m3
[...]

PR #899, opened 2026-07-18
Add BAAI/bge-m3 sparse (lexical_weights) support via --pooling m3_sparse

SpongeRobert:
## Summary

Adds **BAAI/bge-m3 `sparse` (lexical_weights) support** to the candle backend via a new pooling mode `--pooling m3_sparse`, addressing the core ask of #141. Dense bge-m3 already works in TEI (`--pooling cls`); this applies the model's trained `sparse_linear` head over the same hidden states, so the existing `/embed_sparse` endpoint returns bge-m3 sparse vectors with **no wire-format change**.

Closes the sparse half of #141. (Follows up my design comment above.)

## Motivation

bge-m3 sparse is a separately trained `sparse_linear` layer (`Linear(1024→1)`, shipped as `sparse_linear.pt`), **not** an MLM head — so `--pooling splade` can't produce it (`load_roberta` has no `lm_head.decoder`). Today the README/model support reflects "bge-m3, no sparse". This PR closes that gap while reusing the existing SPLADE plumbing.

## Design

- **New `Pool::M3Sparse`** (`backends/core/src/lib.rs`), CLI `--pooling m3_sparse` (kebab alias `m3-sparse` for CLI consistency).
- **`BertM3Head`** (`backends/candle/src/models/bert.rs`): loads `sparse_linear.pt` via `VarBuilder::from_pth` (candle pickle; fp16→f32), computes `relu(sparse_linear(hidden)).squeeze(-1)`, then **scatter-max over the input token ids** into a dense `vocab_size` vector, dropping special tokens and non-positive weights. The router's existing `sparsify` closure turns that into `{index, value}` pairs — **routers/types untouched**. Unlike SPLADE, indices are the *input* token ids (not vocab argmax), so `input_ids` are threaded into pooling.
- **Both device paths**: CPU/Metal (`bert.rs`) and CUDA/flash (`flash_bert.rs`) share one `token_weights` + `scatter_sequence` implementation, so the two paths can't diverge. Flash uses `cumulative_seq_lengths` to scatter per-sequence.
- **`core/src/download.rs`** fetches `sparse_linear.pt` when `pool == M3Sparse`.
- Backends that don't support it (`ort`, `python`) reject `m3_sparse` the same way they reject `Splade`.

## Correctness / parity

Verified **bit-exact against FlagEmbedding** (`BGEM3FlagModel`) on a golden fixture including a Russian sentence:

| path | index-set Jaccard | max \|Δ value\| |
|---|---|---|
| CPU (f32) | **1.0** | ≤ 6.7e-4 |
| CUDA / flash, A100 (fp16) | **1.0** | 6.4e-4 |

The residual ~1e-4 is expected: the golden reference uses transformers' exact GeLU, whereas TEI logs `hidden_act=gelu → GeLU+tanh approximation` — index sets are identical. Test + fixture added under `backends/candle/tests/` (runs in `cargo test`).

## Points I'd like maintainer input on

These are the questions from my design comment, now concrete in code:

1. **Detection** — I went with an **explicit `--pooling m3_sparse`** rather than auto-detecting from `sparse_linear.pt` presence. Happy to switch to auto-detect if you prefer.
2. **Two small public-ish API touch-points** (flagged for review, not hidden):
   - `is_splade()` → renamed **`is_sparse()`** (gates `/embed_sparse`, `embed_all`, `normalize` — semantics shared by both sparse poolings).
   - `download_artifacts(...)` gains a `pool: Option<Pool>` arg (to fetch `sparse_linear.pt`).
   - `Pool` now derives `Copy`.
   I can shape these differently to minimize churn if you'd rather.
3. **Scope** — candle-only, matching where SPLADE lives.

## Notes

- Branch is based on the **`v1.9.3` release tag**; glad to rebase onto `main` on request.
- README updated to list the `m3_sparse` pooling option.

---

_Disclosure: implementation and this PR were prepared with AI assistance; the parity results above were run and verified by me._


SpongeRobert:
Rebased onto current `main` (v1.9.4, `29ccc53`) — the branch now merges cleanly.

**Conflict.** There was one, in `gemma3.rs`: `main` added last-token pooling for Gemma3. I kept that and only added `Pool::M3Sparse` to the unsupported (`unreachable!`) arm, next to CLS and SPLADE. The diff is otherwise unchanged (5 commits, +744/−44, 36 files), and it compiles alongside the models added since v1.9.3 (pplx1, voyage-nano).

**Re-verified after the rebase** (toolchain 1.92.0, per `rust-toolchain.toml`):

- `cargo clippy -- -D warnings`, `cargo fmt --check`, `cargo build -F candle -F grpc --no-default-features`: clean
- `test_m3_sparse`: parity against FlagEmbedding unchanged: identical index sets, max |Δ value| ≤ 6.7e-4 on all golden cases, including the Russian one
- `test_bert` and `text-embeddings-backend-core`: pass

The CI workflows need a maintainer's approval to run on this fork PR. It's ready for review whenever someone has a moment. The open questions from the description still stand: explicit `--pooling m3_sparse` vs auto-detection, and the `is_splade()` → `is_sparse()` / `download_artifacts(…, pool)` changes. I'm happy to reshape either.

PR #916, opened 2026-08-15
fix: remove host-native CPU target

### LINKED ISSUE #902: target-cpu=native causes SIGILL when running AMD-built binaries on Intel CPUs
### Feature request

`.cargo/config.toml` globally sets `target-cpu=native`. 
A binary built on an AMD host generated the SSE4a `EXTRQ` instruction and crashed with SIGILL when deployed on an Intel host without SSE4a.
The issue is triggered when calling the `/embed` endpoint.

### Motivation

Ensure that binaries built on AMD CPUs can run on Intel CPUs.

### Your contribution

Remove the global `target-cpu=native` setting.

atirna:
## What does this PR do?

Removes the repository-wide `target-cpu=native` Rust flag so Linux binaries do not inherit CPU features from the build host. This prevents an AMD-built binary from using AMD-only instructions such as SSE4a when deployed on an Intel CPU.

The target-specific WASM SIMD and x86_64 macOS AVX restrictions remain unchanged.

Fixes #902

## Verification

- Confirmed the config regression check fails before this change because the global `[build]` section exists, then passes after removal while preserving both target-specific flag sets.
- `cargo metadata --no-deps --format-version 1`
- `cargo check -p text-embeddings-router`
- `git diff --check`

`cargo fmt --check` could not run because this checkout does not contain generated gRPC modules (`backends/grpc-client/src/pb.rs` and `router/src/grpc/pb.rs`).

## Before submitting

- [x] I read the contributor guideline.
- [x] This change was discussed in #902.
- [x] No documentation update is needed for this build-configuration-only fix.
- [x] No source-level test is applicable; the configuration regression check and compile verification above cover the change.


atirna:
closing this one out for now, its been sitting a while and i dont want to leave a stale branch on your queue. happy to reopen or rebase it if its still useful to you.

